PR #1, opened 2018-06-14
Removing Json.net Dependency

leniency:
Changed two things:

- added a few .gitignore entries from opening in VS2017.
- Removed Json.net Dependency. It wasn't being used by anything, and was set to v11. This makes running from an Azure Function annoyingly difficult. Azure Functions have some known issues with locked dependencies, Json.Net being one of them (Functions all run directly in the host, which already is locked to Json.Net 9.0.1). Best to just remove the dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,12 @@ project.lock.json
 .idea/
 obj/
 .DS_Store
+
+# User-specific files
+*.suo
+*.user
+*.userosscache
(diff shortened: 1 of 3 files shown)


lukemurray:
Thanks for contributing! Interested to know how you're using this library?

I've actually had had time recently and I was making some changing.

PR #16, opened 2019-06-05
Introspection Query

JTravis76:
Start of Introspection Query. Summary of changes below. Copied of email for tracking.

- Add Models (folder) namespace to avoid collision with the Field object while working with Introspection return model.

- GraphQLIgnoreInputAttribute will remove the parameter from the INPUT type. Ex: might be to hide a virtual ICollection or navigation property.

- SchemaBuilder.cs: pull description from "ComponentModel.DescriptionAttribute".

- SchemaType.cs: pull description from "ComponentModel.DescriptionAttribute" & changed string.Empty to null. This will allow a new Field to update the ReturnTypeSingle.

- Update Demo with typos, descriptions, and checking for Introspection Query

### DIFF
--- a/src/EntityGraphQL/Compiler/GraphQLResultNode.cs
+++ b/src/EntityGraphQL/Compiler/GraphQLResultNode.cs
@@ -67,6 +67,12 @@ namespace EntityGraphQL.Compiler
                 var data = node.Execute(context);
                 result.Data[node.Name] = data;
             }
+
+            if (GraphQLVaildation.Errors.Count > 0)
+            {
+                result.Errors.AddRange(GraphQLVaildation.Errors);
+            }
+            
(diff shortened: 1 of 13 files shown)


lukemurray:
Thanks for this. Most of it is great, the one thing I don't love is the change in the Query controller to look for a specific named query to trigger the introspection query.

GraphQL is more generic than that, we should be able to write any query and name it anything, see https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsIntrospection-test.js for examples and https://graphql.org/learn/introspection/

Next week when I get a chance I'll merge this into a branch and work on some changes so that any query can query the metadata (just like `__typename`).

Thanks heaps for starting this!

PR #27, opened 2019-09-09
Feature/sonarfixes

driekus77:
Hi All,

Found your repository promising and looked at it if I can help. Don't know much yet about GraphQL but willing to learn. 

I scanned the code using SonarLint and SonarQube and found some issues that I tried to fix in this PR.

Also discovered, while debugging, that code is unstable: Depending on the location of the Break point(s) the unit test(s) outcome where different. Maybe Garbage Collection kicks in to soon for the DynamicTypes?

Maybe this PR can be used as reference for fixing to instability issues?

Keep up the good work!

Kind regards,
Henry


lukemurray:
Thanks for this. There is a lot here, big one is I don't want to bring in the files under `.sonarlint` especially since they have hard coded paths in them.

It also changed all white space from spaces to tabs so it makes the PR hard to follow the real changes. Any change you can resubmit without those files and without changing white space?

Thanks

PR #29, opened 2019-10-28
split ignore into query and input

### LINKED ISSUE #20: Separate GraphQLIgnore attribute into input and query
If you have an object that you share across a query object and as an input into a mutation you may want to label some as ignore from query and others ignore from the mutation input.

As an alternative right now, you can use simple DTOs for mutation inputs.

lukemurray:
fixes #20 

Use `[GraphQLIgnore]` to mark a field as ignored in both Input types and Query schema in EntityGraphQL.

Use `[GraphQLIgnore(GraphQLIgnoreType.Query)]` to mark a field as ignored in the query schema. If the type is also used in a mutation input type, the field is available.

Use `[GraphQLIgnore(GraphQLIgnoreType.Input)]` to ignore a field in a mutation input type if the type is also exposed in the query schema, it will be available.

Note `[GraphQLIgnore]` is shorthand for `[GraphQLIgnore(GraphQLIgnoreType.All)]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
-# 0.21.1
+# 0.22.0
+- You can now specify if a field is ignored for queries, mutations or both (if you're sharing DTOs/objects)
 - Don't output the meta information of the schema in the schema definition
 - Prevent duplicate scalar types in schema generation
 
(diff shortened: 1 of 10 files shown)


lukemurray:
@JTravis76, does this solve the issue of being able to ignore fields in mutations vs queries?

PR #30, opened 2019-12-23
Model Validation - Mutation

JTravis76:
## Model Binding Validation  
Goal is to provide feedback to the end-user when posting a mutation that fails to provide a required field.

**Example Usages:**  
_Add Required attribute to field(s) that are required._  
> ErrorMessage are optional. When omitted, one is generated
```cs
using System.ComponentModel.DataAnnotations;

public partial class Group {
    [Key]
    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
    [Column(TypeName = "BIGINT")]
    public long Id { get; set; }

    [Required(AllowEmptyStrings = false, ErrorMessage = "Group Name is required")]
    [StringLength(50)]
    [Column(TypeName = "VARCHAR(50)")]
    public string GroupName { get; set; }

    [Column(TypeName = "BIGINT")]
    [System.ComponentModel.Description("Foreign Key to Group Type")]
    public long? GroupTypeId { get; set; }
}
```
_In mutation method, add validation error and/or check for existing errors._
```cs
public class Mutations
{
    private readonly MappedSchemaProvider<DataContext> _schema;

    public Mutations(MappedSchemaProvider<DataContext> schema)
    {
        _schema = schema ?? throw new ArgumentNullException(nameof(schema));
    }

    [GraphQLMutation("Add/Update an Email Group")]
    public Group AddUpdateGroup(DataContext context, Group group)
    {
        //== Manually check if foreign key is provide, else add validation error
        if (group.GroupTypeId == null || group.GroupTypeId <= 0)
            _schema.AddError("Group Type Id is required");

        using DataContext db = context;

        if (group.GroupId == 0)
             db.Add(group);
        else
            db.Update(group);

        //== No validation errors, then save changes
        if (_schema.IsVaild)
            db.SaveChanges();

        return group;
    }
}
```
_(optional) Send a `Bad Request` Code (400)_
```cs
[AllowAnonymous]
[HttpPost]
public object Post([FromBody]QueryRequest graph)
{
    QueryResult data = _context.QueryObject(graph, _schemaProvider);

    // This work in GraphiQL while also working with axios http status code handling
    if (data.Errors.Count > 0)
            return BadRequest(new { errors = data.Errors });

    return Ok(data);
}
```


src/EntityGraphQL/Schema/MutationType.cs
```diff
@@ -176,5 +177,46 @@ public ArgType GetArgumentType(string argName)
...
+        /// <param name="entity"></param>
+        private void VaildateModelBinding(object entity)
+        {
+            Type argType = entity.GetType();
+
+            foreach (var prop in argType.GetProperties())
+            {
+                object value = prop.GetValue(entity, null);
+
+                //Did this way so we won't have to reference the DLL
```
lukemurray: Is it an issue if we reference DLL?

lukemurray:
Thanks for this. Looks fair simple and straight forward. 

What are your thoughts on the errors/non-valid schema throwing an exception vs not? Currently in mutations I do things like 
```c#
            if (string.IsNullOrEmpty(Name))
                throw new ArgumentException("Name must not be empty");
```

src/EntityGraphQL/Schema/MutationType.cs
JTravis76: No real concern, just thought you may want to keep the dependencies to a minimum. Whichever you feel is the best.

JTravis76:
Had to do a quick test here to understand before I commented. `throw Exception` will work as expected and display message in GraphiQL. But while in development/ debugging mode, it forces the code to haul. It fine in a Release build. So mainly a personal preference here. I prefer not to be brother by VS unless something is really wrong. 

Also, not sure how else to validate before using the `db.SaveChanges()` for existing errors.

Most of this design models after Microsoft's ModelState object. Definitely a good discussion.

JTravis76:
> Thanks for this. Looks fair simple and straight forward.
> What are your thoughts on the errors/non-valid schema throwing an exception vs not? Currently in mutations I do things like
>             if (string.IsNullOrEmpty(Name))
>                 throw new ArgumentException("Name must not be empty");

Main thought on using the .AddError() was to build a collection of validation errors vs. a single one-off notice (which could be used also). I like to display ALL validation errors for a user to correct. Example: When filling out of form, upon saving, it tell me that Name is required, I add name and hit save again, then it tell me that Address is required. Personally, rather display both Name and Address is required upon saving.

lukemurray:
Are you closing this to create a new PR? This is still on my list to visit,
it just hasn't been the top of my list.

JTravis76:
Sorry,  about that. I was bringing my forked branch up to date with your
latest commits. And I forgot about the pull request. I will merge up the
changes from my dev branch and resubmit.

Brain failure on my part :)

lukemurray:
all good, the latest changes have breaking chnages let me know what you
think, read the changeling and the readme changes, I'm trying to make it
clearer

PR #39, opened 2020-04-11
Model Validation - Mutation

JTravis76:
Goal is to provide feedback to the end-user when posting a mutation that fails to provide a required field.  
Developers could throw an error which is caught by EntityGraphQL and pass into the error property of the QueryResult object. But is only provides a single validation, not a collection.
```cs
if (group.GroupTypeId == null || group.GroupTypeId <= 0)
    throw new ArgumentException ("Group Type Id is required");
```

**Example Usages:**  
_Add Required attribute to field(s) that are required._  
> ErrorMessage are optional. When omitted, one is generated
```cs
using System.ComponentModel.DataAnnotations;

public partial class Group {
    [Key]
    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
    [Column(TypeName = "BIGINT")]
    public long Id { get; set; }

    [Required(AllowEmptyStrings = false, ErrorMessage = "Group Name is required")]
    [StringLength(50)]
    [Column(TypeName = "VARCHAR(50)")]
    public string GroupName { get; set; }

    [Column(TypeName = "BIGINT")]
    [System.ComponentModel.Description("Foreign Key to Group Type")]
    public long? GroupTypeId { get; set; }
}
```
_In mutation method, add validation error and/or check for existing errors._
```cs
public class Mutations
{
    private readonly SchemaProvider<DataContext> _schema;

    public Mutations(SchemaProvider<DataContext> schema)
    {
        _schema = schema ?? throw new ArgumentNullException(nameof(schema));
    }

    [GraphQLMutation("Add/Update an Email Group")]
    public Group AddUpdateGroup(DataContext context, Group group)
    {
        //== Manually check if foreign key is provide, else add validation error
        if (group.GroupTypeId == null || group.GroupTypeId <= 0)
            _schema.AddError("Group Type Id is required");

        if (group.GroupId == 0)
             context.Add(group);
        else
            context.Update(group);

        //== No validation errors, then save changes
        if (_schema.IsVaild)
            context.SaveChanges();

        return group;
    }
}
```
_(optional) Send a `Bad Request` Code (400)_
```cs
[AllowAnonymous]
[HttpPost]
public object Post([FromBody]QueryRequest gql)
{
    QueryResult data = _schemaProvider.ExecuteQuery(gql, _dataContext);

    // This work in GraphiQL while also working with axios http status code handling
    if (data.Errors.Count > 0)
            return BadRequest(new { errors = data.Errors });

    return Ok(data);
}
```


lukemurray:
Thanks, I'll try to have a look at it this coming week.

JTravis76:
No rush, I pushed it from another branch so I may refresh master without
breaking it :)

Jeremy

lukemurray:
The main code in `VaildateModelBinding` looks good and the general idea is great.

I think instead of making people pass the schema into the mutation classes we can use the new DI functionality.

The library could register an `ErrorProvider` or something in the services and if you need to use it in your mutation method you just list and and it will be resolved. And the Execute query code can read that out.

If you don't have the time I could make that change and include it in a refactor I have coming that also adds directives.

JTravis76:
Sure, go ahead. Be interested in see how the DI works out. I notice that new feature to inject services. Seems nice.

lukemurray:
Thanks for this, see the updated validation section in the readme to see how I adapted it with the service style.

PR #57, opened 2020-06-05
Escape " in Description field when generating schema.

AnderssonPeter:
When the description contains a " the generated schema is invalid, this pull request replaces " with \\\" in the generated schema.


AnderssonPeter:
after thinking about it, maybe we should escape the \ char also.

PR #64, opened 2020-09-15
GetField by an expression

AradAral:
Get a field by an expression selection on the real type.
A useful overload for the `GetField(string fieldName)` method. It takes an `Expression<Func<TBaseType, object>>` rather than a `string`.
It works just like the overload for the `RemoveField(string fieldName)` method, which currently exists, that method too takes an expression rather than a string and then does the work, which makes our lives easier.


AmirHosseinHmd:
I forgot to return the field! :D
I fixed it.

lukemurray:
Thanks!

PR #71, opened 2020-10-07
Defects/error array

GravlLift:
Fix the issue where an empty error array is sent with every response.

`errors` property should not be present on the response if there are no errors per the [graphQL specification](https://spec.graphql.org/June2018/#sec-Errors).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,10 @@ obj/
 # Visual Studio 2015 cache/options directory
 .vs/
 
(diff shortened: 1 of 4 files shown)


lukemurray:
Thanks!

lukemurray:
I also updated the tests in master and bumped the version. I have a couple of things I want to look at before pushing a new nugget package though. I'll push this soon if I don't get a chance to look at the other things.

PR #74, opened 2020-10-13
Use FieldNamer instead of SchemaGenerator.ToCamelCaseStartsLower.

AnderssonPeter:
Currently the `ToCamelCaseStartsLower` is used in some cases and the `FieldNamer` in others, this desync causes errors for example with mutation arguments, giving the error that a field is not found.

I fixed all the easy ones in this pull request, but there are 5 left that don't operate on `MemberInfo` that i have no idea how to fix.

* SchemaGenerator.BuildMutations
* SchemaGenerator.BuildSchemaTypes
* SchemaGenerator.GetGqlArgs
* SchemaGenerator.MakeQueryType
* SchemaIntrospection.BuildFieldsForType

### DIFF
--- a/src/EntityGraphQL/Schema/MutationType.cs
+++ b/src/EntityGraphQL/Schema/MutationType.cs
@@ -12,6 +12,7 @@ namespace EntityGraphQL.Schema
 {
     public class MutationType : IField
     {
+        private readonly ISchemaProvider schema;
         private readonly object mutationClassInstance;
         private readonly MethodInfo method;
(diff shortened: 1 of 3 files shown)


lukemurray:
Thanks @AnderssonPeter, good pick up!

lukemurray:
Probably best we have the renamed just take a string and return a string.

lukemurray:
Thanks, I moved it to `Func<string, string>`. Also found a lot more that was not using it! Will do a bit more testing before releasing, let me. know if you hit anything else related to this. Thanks

PR #75, opened 2020-10-13
Replaced GraphQLNotNullAttribute with RequiredAttribute.

### LINKED ISSUE #69: Inconsistent usage of attributes to mark a field as not null/required?
There might be a good reason for this if that is the case just close the issue!

But when marking a argument as required in some parts of this library use the `RequiredAttribute` and some parts use the `GraphQLNotNullAttribute` attribute.

I was confused why applying the `RequiredAttribute` to a field on my Entity had no effect, after looking at the source i found the `GraphQLNotNullAttribute` applying that attribute ensured that the field was marked as required.

AnderssonPeter:
Resolves #69


lukemurray:
Thanks @AnderssonPeter.

I'd like to try and start making efforts at preventing breaking changes. So it'd be good to mark the `GraphQLNotNullAttribute` as deprecated and still have a helper method that checks for either `GraphQLNotNullAttribute` or `RequiredAttribute`.

In a later version we can clean up the deprecations. I know I've made breaking changes many times, but trying hard now to make that smoother! This one should be fairly straight forward too.

lukemurray:
Fixed in 1.1

PR #77, opened 2020-10-16
Added string and long parsing for DateTime and DateTimeOffset

GravlLift:
Added string and long parsing for DateTime and DateTimeOffset. In my particular use case, I had a field that accepted a `DateTimeOffset` as a parameter. This change allows for the ability to pass that parameter as either a formatted string, or ticks.

```
{
  entity(startDate: 637384711224902655, endDate: "2020-10-16T18:54:19.5109301-04:00") {
    id
    latitude
    longtitude
    address1
    city
    country
    postalCode
    state
  }
}
```


lukemurray:
Thanks!

Issue #96, opened 2022-02-10
Type int inside InputType list excepted to have long instead of int


pierrebelin:
Hello!

I'm trying to update lists of ids through mutation.
But I have an error on the `int` type: Object of type 'System.Int64' cannot be converted to type 'System.Int32'

```
[MutationArguments]
public class UserGroupCreateArgs 
{
    [Required(AllowEmptyStrings = false, ErrorMessage = "Display name is required")]
    public string DisplayName { get; set; }
    public List<ListIdArg> Users { get; set; }
}

public class ListIdArg
{
    public int Id { get; set; }
}
```

This is my GraphQL query. I'm trying to fill the list `users` with a list of `int`.

```
mutation { CreateOrUpdateUserNotificationGroup(
  id:100
  displayName: "TestName", 
  users:[{id:1}])
  { id } 
}
```

But I've got this message.

```
{
  "data": {},
  "errors": [
    {
      "message": "Object of type 'System.Int64' cannot be converted to type 'System.Int32'."
    }
  ]
}
```

Since `id` does throw the error, but only the `id` inside the list `user`, I guess it's related to the fix #92 which fixed the list as input

I bypass the problem by changing the type `id` inside `ListIdArg` from `int` to `long`.

It's a bit annoying because it needs to be converted after, but not an hard issue.

PR #97, opened 2022-02-11
feat(mutation): Handle long and int as argType

pierrebelin:


pierrebelin:
It comes to fix the issue #96

lukemurray:
Thanks!

PR #102, opened 2022-03-22
fix: Handling dictionaries introspection

pierrebelin:
Hi
This is not a big deal, but I need to manipulate dictionary through GraphQL, and I get this error in introspection:
`{"data":{},"errors":[{"message":"Field error: __schema - Value cannot be null. (Parameter 'type')"}]}`

GraphQL is working, but this is painfull to not have introspection to develop the front.

Regards

### DIFF
--- a/src/EntityGraphQL/Extensions/TypeExtensions.cs
+++ b/src/EntityGraphQL/Extensions/TypeExtensions.cs
@@ -80,6 +80,8 @@ namespace EntityGraphQL.Extensions
                 return type.GetElementType();
             if (type.GenericTypeArguments.Count() == 1)
                 return type.GetGenericArguments()[0];
+            if (type.GenericTypeArguments.Count() == 2)
+                return type.GetGenericArguments()[1];
             return null;
         }
 
--- a/src/tests/EntityGraphQL.Tests/TestDataContext.cs
+++ b/src/tests/EntityGraphQL.Tests/TestDataContext.cs
@@ -19,6 +19,7 @@ namespace EntityGraphQL.Tests
         public List<Location> Locations { get; set; } = new List<Location>();
         public virtual List<Person> People { get; set; } = new List<Person>();
         public IEnumerable<User> Users { get; set; } = new List<User>();
+        public Dictionary<string, object> Attributes { get; set; } = new Dictionary<string, object>();
     }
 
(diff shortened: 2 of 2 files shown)


lukemurray:
Thanks for this. I ended up changing it slightly. This made it not fail on building the schema but if you tried to query it, it would then fail.

As the GraphQL type system doesn't have a `Dictionary` or `KeyValuePair` type built in (or generics). The way to do this (will be in 1.3.0) is by first telling the schema about the mapping

```c#
services.AddGraphQLSchema<DemoContext>(builder =>
{
    builder.PreBuildSchemaFromContext = schema =>
    {
        // add in needed mappings for our context
        schema.AddScalarType<KeyValuePair<string, string>>("StringKeyValuePair", "Represents a pair of strings");
    };
});
``` 

Then the schema generated is valid

```
scalar StringKeyValuePair
type Query {
	attributes: [StringKeyValuePair!]
}
```

And you can query it and get results as `[{key: value}]`

If you only want it to return the values (not keys) you can modify the schema.

```c#
services.AddGraphQLSchema<DemoContext>(builder =>
{
    builder.PreBuildSchemaFromContext = schema =>
    {
        // add in needed mappings for our context
        schema.AddScalarType<KeyValuePair<string, string>>("StringKeyValuePair", "Represents a pair of strings");
    };
    builder.ConfigureSchema = schema => 
    {
        schema.ReplaceField("attributes", cox => ctx.Attributes.Values, "Attribute values");
    }
});
```

lukemurray:
Small change to above when it gets published is that you can still use `builder.PreBuildSchemaFromContext` if you want to control name and description but if you leave it out by default the schema builder will create those key value pairs.

pierrebelin:
Exactly, it can be an huge improvment.

This is how I use it actually :

Declaring type :
`schemaProvider.AddType<Dictionary<string, object>>("dictionaryStringObject", "Dictionnary string/objects");`

For the property :
public Dictionary<string, object> sums { get; set; }

Then I add the field (I use a custom AddField, but the main thing is there)
`AddField("sums", schemaType, typeof(Dictionary<string,object>), "sums", $"Sums of {entityType.Name}");`

I can then query sums, but I'll get all keys. In my case, this is exactly what I want.

PR #104, opened 2022-03-30
fix(argument): Handling float/double/decimal with dot decimal

pierrebelin:
Test FloatArg in ArgumentTests.cs was no longer working.


lukemurray:
Hi, thanks. I'm doing a bit of changes for a 2.0 release. It already has. the `Dictionary<>` support (as mentioned you'll need to override the default if you do not want key, value and only values).

Can you add some detail of where this is failing to correct parse a float/decimal such as `"3.2"`? I use decimals throughout projects fine. 

Test `FloatArg` is working here. Is this a different language platform you are running?

Thanks

lukemurray:
Hi, thanks. I'm doing a bit of changes for a 2.0 release. It already has. the `Dictionary<>` support (as mentioned you'll need to override the default if you do not want key, value and only values).

Can you add some detail of where this is failing to correct parse a float/decimal such as `"3.2"`? I use decimals throughout projects fine. 

Test `FloatArg` is working here. Is this a different language platform you are running?

Thanks

lukemurray:
Hi, this now conflicts with the 2.0 changes. Can you help me test things with your CultureInfo? Is it `CultureInfo.InvariantCulture` that helps or both that and `NumberStyles.AllowDecimalPoint`? As you do not use decimal p point right?

pierrebelin:
Of course, I can take a look this weekend.
Both are helping : `CultureInfo.InvariantCulture` means that I'll parse independently of local settings (Fr-fr/En-en...), so I can had rules to my parser, here `NumberStyles.AllowDecimalPoint`. It's better to specify it, because it forces string to be written with a `.` instead of an `,`

pierrebelin:
@lukemurray I fixed the parsing problems, `CultureInfo.InvariantCulture` resolve the problem without `NumberStyles.AllowDecimalPoint` 
I also fixed some tests validation to have successful checks

PR #111, opened 2022-05-04
Attempt at adding interfaces

bzbetty:
Not quite working yet - have an issue where __typename returns the base type instead of the object type itself.  But figured I'd start the chat to see if I'm on the right track.

`AddField("__typename", t => name, "Type name").IsNullable(false);`

is obviously always going to show the base type.

`AddField("__typename", t => schema.Type(t!.GetType().Name).Name, "Type name").IsNullable(false);`

Works for most the tests, but seems to break on getting typename on a relationship (as it's coming through as IEnumerable

### DIFF
--- a/EntityGraphQL.sln
+++ b/EntityGraphQL.sln
@@ -5,19 +5,19 @@ VisualStudioVersion = 15.0.26124.0
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{7214E34E-E71C-4604-9FD4-7ADDDFA09F26}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "EntityGraphQL", "src\EntityGraphQL\EntityGraphQL.csproj", "{5D8B02DC-20EF-456F-A1DB-840FF1455156}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EntityGraphQL", "src\EntityGraphQL\EntityGraphQL.csproj", "{5D8B02DC-20EF-456F-A1DB-840FF1455156}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{5F132C01-AA1F-4241-B90F-BE4B2275B34E}"
 EndProject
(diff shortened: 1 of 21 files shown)


lukemurray:
Hi, thanks for this!

Without digging `AddField("__typename", t => schema.Type(t!.GetType().Name).Name, "Type name").IsNullable(false);` looks good. You may have to check if `t!.GetType()` is an enumerable or array but based on how types are registered I don't think that should be required.

I might have to have a play with this myself. I assume the failing tests are in this PR? I'll check it out and have a look in a few days when I get a chance.

bzbetty:
The current failing test is TestInheritance in QueryTests.cs

If I change the code you mention above:

* TestInheritance then passes
* TestCollectionCollectionObjectThenService in ServiceFieldTests.cs then fails

`Field 'projects' - variable 't' of type 'EntityGraphQL.Tests.Person' referenced from scope '', but it is not defined`

which to me sounds like the expression parameter was somehow removed

bzbetty:
ok pushed a new version that removes the special case to not rebind __typename, all tests seem to pass.  Not sure why that special case existed though so potentially broke something else.

lukemurray:
Ahh nice. I'd like to confirm, but I think you change is correct. Previously it was assuming the `__typename` field never needed to be updated on an execution with service fields. But the change you have to the `__typename` field - which is correct to support interfaces - requires that code (if the if statement) to run as the we have changed the type `t`.

Very nice.

lukemurray:
Be awesome if you could have a go at updating the docs too :)

bzbetty:
First take at documentation added.


I also tried having a look at how you could support

```
query {
    animals {
        __typename
        name
        ... on Dog {
            hasBone
        }
    }
}
```

figured you'd need to override VisitInlineFragment and switch the context but couldn't get it working.

lukemurray:
Certainly will need to visit inline fragment.

Then when the GraphQl*Nodes are putting together the expressions it likely has to build something like
```
type is Dog ? a.hasBone : null
```
for field `hasBone`

Which will be interesting as that field will be in each object in the result list (with null value). Unless the conditional is at the select level.

```
// poor example
animals = animals.Select(a => a is Dog ? new { a.name, a.hasBone } :  new { a.name }
```

Not sure if that helps, but I'd need to think about it more too.

lukemurray:
Bring this in :) can look at adding inline frag support separately. Thanks.

bzbetty:
Don't believe anything is missing, ultimately i didn't really find it useful without fragments, but if it's a step in the right direction that's good.

lukemurray:
Fair point regarding it not being super useful with inline frags... might be worth add that together then.

bzbetty:
Thinking further, this would still be useful in tph scenarios. Might still be worth merge

PR #118, opened 2022-06-04
Wildcard queries don't honor removed fields

breyed:
When a query does not include any fields, all fields are returned even if some fields have been removed via `RemoveField`.

For example, even with this configuration:
``` c#
schema.Type<Person>().RemoveField(p => p.Id);
```
the query `{ people }` returns the `Id` field.

This behavior limits the ability for EntityGraphQL to be used with entity types used internally by the server. It would be convenient to add app's existing entity types to a GraphQL DB context, except with sensitive and incompatible fields removed. However, since a GraphQL client can access all fields, regardless of `RemoveField`, this approach isn't currently viable.

To reproduce this issue, run the test `WildcardQueriesHonorRemovedFields` in this PR.

### DIFF
--- a/src/EntityGraphQL.AspNet/PolicyOrRoleBasedAuthorization.cs
+++ b/src/EntityGraphQL.AspNet/PolicyOrRoleBasedAuthorization.cs
@@ -30,7 +30,7 @@ namespace EntityGraphQL.AspNet
             // if the list is empty it means identity.IsAuthenticated needs to be true, if full it requires certain authorization
             if (requiredAuth != null && requiredAuth.Any() && user != null)
             {
-                // check polcies if principal with used
+                // check polices if principal with used
                 if (authService != null)
                 {
                     var allPoliciesValid = true;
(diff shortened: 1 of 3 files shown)


lukemurray:
oh thanks for the pick up and making a unit test.

lukemurray:
Reading the GraphQL [spec](https://spec.graphql.org/October2021/#sec-Field-Selections) (also the invalid queries [here](https://spec.graphql.org/October2021/#sec-Leaf-Field-Selections)), it seems selection sets are required. Suggesting we should throw an error.

> Field selections must exist on Object, Interface, and Union types

Makes sense as the goal is clients being specific about the data/fields they are requesting. Wondering if you are hoping to use empty selection sets on lists of object fields or the main issue is that you _can_ which can expose unwanted data?

breyed:
Throwing an error makes sense to me. We didn't have any plans to use a wildcard in production. I noticed that it (kind of) worked and so was using it for experimenting. When the field wasn't removed, it had me concerned there could be other problems I wasn't aware of. Disallowing wildcards in the first place would avoid the confusion.

PR #121, opened 2022-06-06
Add default mappings for 8-bit integer types.

breyed:
These are useful for mapping to 8-bit database columns.


lukemurray:
This makes sense. I just want to check how `byte[]` will be handled. We use the default serialisation which puts it to a string. But I think this will push it to an array of ints. 

These can always be added in the projects own schema of course.

lukemurray:
This makes sense. I just want to check how `byte[]` will be handled. We use the default serialisation which puts it to a string. But I think this will push it to an array of ints. 

These can always be added in the projects own schema of course.

lukemurray:
Ha! I missed that the `byte[]` mapping is defined in EntityGraphQL!!

PR #128, opened 2022-06-10
Grammar and redundancy cleanup

breyed:
Various grammar and typo fixes. Also removed the bit about arrays from the context that had mentioned both arrays and IEnumerable<T>, since all arrays already implement IEnumerable<T>.


lukemurray:
Thanks for these!

PR #129, opened 2022-06-12
add basic support for nullable ref types

bzbetty:
given the following context

```
public class Context
{
   public string NonNullable { get; set; } = "";
   public string? Nullable { get; set; }   
   public IEnumerable<Test> Tests { get; set; } = new List<Test>();
   public IEnumerable<Test>? Tests2 { get; set; }
}
```

generate the following query type

```
type Query {
	nonNullable: String!
	nullable: String
	tests: [Test!]!
	tests2: [Test!]
}
```

### DIFF
--- a/src/EntityGraphQL/Schema/GqlTypeInfo.cs
+++ b/src/EntityGraphQL/Schema/GqlTypeInfo.cs
@@ -16,13 +16,13 @@ namespace EntityGraphQL.Schema
         /// <param name="schemaTypeGetter">Func to get the ISchemaType. Lookup is func as the type might be added later. It is cached after first look up</param>
         /// <param name="typeDotnet">The dotnet type as it is. E.g. the List<T> etc. </param>
         /// <param name="nullableValueTypes">value types are nullable. Used for arguments where they may have default values</param>
-        public GqlTypeInfo(Func<ISchemaType> schemaTypeGetter, Type typeDotnet, bool nullableValueTypes = false)
+        public GqlTypeInfo(Func<ISchemaType> schemaTypeGetter, Type typeDotnet, bool nullableValueTypes = false, bool nullableReferenceType = true)
         {
             SchemaTypeGetter = schemaTypeGetter;
 
             TypeDotnet = typeDotnet;
-            TypeNotNullable = !nullableValueTypes && TypeDotnet.IsValueType && !TypeDotnet.IsNullableType();
             IsList = TypeDotnet.IsEnumerableOrArray();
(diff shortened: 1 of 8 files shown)


src/EntityGraphQL/Schema/GqlTypeInfo.cs
```diff
@@ -16,13 +16,13 @@ public class GqlTypeInfo
         /// <param name="schemaTypeGetter">Func to get the ISchemaType. Lookup is func as the type might be added later. It is cached after first look up</param>
         /// <param name="typeDotnet">The dotnet type as it is. E.g. the List<T> etc. </param>
         /// <param name="nullableValueTypes">value types are nullable. Used for arguments where they may have default values</param>
-        public GqlTypeInfo(Func<ISchemaType> schemaTypeGetter, Type typeDotnet, bool nullableValueTypes = false)
+        public GqlTypeInfo(Func<ISchemaType> schemaTypeGetter, Type typeDotnet, bool nullableValueTypes = false, bool nullableReferenceType = true)
```
lukemurray, line 19: Do we need to know they are nullable value or reference types?

`nullableValueTypes` might have been misleading, from memory it is just stating that the "field value" is nullable or not vs. the element value of a list.

bzbetty: It would probably make sense to merge those two parameters into a single isnullable

bzbetty: So couple of bits here

* I couldn't find anything that set nullableValueTypes to the non default value
* The nullableReferenceType variable would probably be better written as nullableType as it works for int? as well and class? 
*  however given the detection method it has to be detected when you have the PropertyInfo object still which this constructor didn't.
* This means the nullableValueTypes flag enables support for nullableValueTypes, but the nullableReferenceType is stating the passed type is nullable, so there is a bit of a difference there
* The code only added support for nullable ref types at the top level (not sub fields, or manually added fields) - no reason why, would probably be simple enough.

bzbetty: updated
* Passes info MemberInfo (Field, Property or Method) now instead 
* cleaned up nullable checking a little bit
* Now supports mutations and AddAllFields too

lukemurray:
Looks good. Any chance you can resolve the conflicts?

bzbetty:
> Looks good. Any chance you can resolve the conflicts?

Done

PR #132, opened 2022-06-13
Only get types for public instance properties of mutations.

breyed:
There are a few other calls to `GetProperties` in the codebase without arguments. It's not clear whether they also should filter their output.


lukemurray:
Have you hit an issue with this? According to [this](https://docs.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=net-6.0#system-type-getproperties)
> An array of [PropertyInfo](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo?view=net-6.0) objects representing all public properties of the current [Type](https://docs.microsoft.com/en-us/dotnet/api/system.type?view=net-6.0).

It already only returned public properties.

breyed:
The public aspect of the parameterless overload is fine. The problem is that it includes static properties. I was exposing a type through GraphQL that had a static property and got an error that the type of the static property hadn't been added.

Type type being exposed is generated by Protocol Buffers (for data exchange with other systems). The static property is `PropertyDescriptor`, which shouldn't be exposed through GraphQL.

lukemurray:
Got it. That makes sense.

PR #134, opened 2022-06-15
Inline Fragments for interfaces (attempt #2)

bzbetty:
Different attempt to https://github.com/EntityGraphQL/EntityGraphQL/pull/133

This one uses expression trees to try get around the issue mentioned in other PR

Sadly it doesn't actually get around it because the dynamic type it returns still has all properties (can't use multiple dynamic types as it throws exception about different return types).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,33 @@
+# 2.3.0
+
+Changes:
+
+- `AddMutationsFrom` now uses the `ServiceProvider` instance to create the mutation class allowing dependency injection at the contructor level like Controllers.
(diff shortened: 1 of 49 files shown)


bzbetty:
update: this now gets around the extra property issue by creating multiple types that inherit dynamically at runtime.

lukemurray:
Cool. Likely have time to look at this (and others) on the weekend.

I was going to suggest dynamic inheritance or creating a weird "union" type.

Shall I focus on looking at this one and forget attempt 1?

Thanks

bzbetty:
yeah forget attempt 1, it's simpler but overall don't think it's as good.

however there are a couple of hacks in #2 that might need more cleaning up (ParameterReplacer.cs)

Issue #126, opened 2022-06-10
Instance per request for mutation class


breyed:
Currently for mutations, a single instance of the mutations class (e.g. `PeopleMutations`) is created at startup. This differs ASP.NET Core controller classes, for which a new instance is created for each request. The instance-per-request approach allows dependency injection to be performed in the constructor, which is the [preferred approach](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-6.0):
> Prefer requesting dependencies as constructor parameters over resolving services from RequestServices. Requesting dependencies as constructor parameters yields classes that are easier to test.

Additionally, if several methods use the same DI parameters, the code for setting them up can be consolidated in the constructor, rather than having to be duplicated in each request method. EntityGraphQL mutations parallel controller methods and would benefit similarly, especially mutations that perform more than just database operations.

Relatedly, it's worth noting that the DI conventions for controller methods and GraphQL mutations are currently reversed. For controller methods, DI services are [annotated with FromServices](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/dependency-injection?view=aspnetcore-6.0#action-injection-with-fromservices) and request data is not annotated. For GraphQL, DI services are not annotated and mutation arguments are annotated with `MutationArguments`.

It would increase consistency and reduce the learning curve of EntityGraphQL if mutations followed the conventions as MVC controllers for instance-per-request and DI. It would also open the door to not requiring a separate args class for each mutation and instead allowing mutation fields to be specified as parameters, since all parameters not marked with `FromServices` could be considered mutation fields.

lukemurray:
Thanks for the feedback.

1. Mutation class instance per request makes sense to support the common set ups in the constructor as you point out. Today you can have other setup methods in the class to help with that.
2. Happy to add support for `FromServicesAttribute` however in ASP.NET 7 they are making that [optional](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-2/#infer-api-controller-action-parameters-that-come-from-services). I would seek to do the same and optionally support `[FromServices]`
3. Totally agree with making the arguments on the mutation method the top level arguments in the GraphQL schema instead of the argument classes. Been on my mind for some time. `[FromServices]` is not required for this, just need to look up what are services and what isn't

Thanks for the feedback, this is great.

bzbetty:
[...]

PR #136, opened 2022-06-18
Instantiate mutation type when executing it

bzbetty:
potential fix for #126 

Instantiate mutation type when executing it, first via serviceProvider then fallback to Activator.createInstance

also allow registering all mutations from multiple types that match a generic parameter (eg marker interface)

### DIFF
--- a/src/EntityGraphQL/Schema/ISchemaProvider.cs
+++ b/src/EntityGraphQL/Schema/ISchemaProvider.cs
@@ -27,7 +27,7 @@ namespace EntityGraphQL.Schema
         ISchemaType AddInterface(Type type, string name, string description);
         SchemaType<TBaseType> AddInputType<TBaseType>(string name, string? description);
         ISchemaType AddInputType(Type type, string name, string? description);
-        void AddMutationsFrom<TType>(bool autoAddInputTypes = false);
+        void AddMutationsFrom<TType>(TType? mutationClassInstance =  null, bool autoAddInputTypes = false, bool addNonAttributedMethods = false) where TType : class;
         ISchemaType AddScalarType(Type clrType, string gqlTypeName, string? description);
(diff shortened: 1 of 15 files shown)


bzbetty:
second change in this PR that makes GraphQLMutationAttribute optional (controversial change?) will take any public method, static or instance that is declared directly on the type (not inherited).

src/EntityGraphQL/Schema/ISchemaProvider.cs
```diff
@@ -27,7 +27,7 @@ public interface ISchemaProvider
         ISchemaType AddInterface(Type type, string name, string description);
         SchemaType<TBaseType> AddInputType<TBaseType>(string name, string? description);
         ISchemaType AddInputType(Type type, string name, string? description);
-        void AddMutationsFrom<TType>(TType mutationClassInstance, bool autoAddInputTypes = false) where TType : notnull;
+        void AddMutationsFrom<TType>(bool autoAddInputTypes = false);
```
lukemurray, line 30: I try (my best) to avoid breaking changes to the API - trying to follow semver where x.y.z is x - breaking changes, y new features, z fixes. 

Given the change, we could also leave the old methods there and mark them as deprecated. The methods can just do nothing with the object passed in. This wold be a 2.2 release.

Then we can later do a 3 that cleans them up and potentially wraps up more breaking changes.

Thoughts?

bzbetty: Would it work if the mutationClassInstance argument remained but defaults to null?

bzbetty:
new commit
* Add mutationClassInstance parameter back in to not break API - also allows user to pass in the mutation class instance if they prefer doing it that way
* added new parameter addNonAttributedMethods which defaults to false to allow methods not marked as mutations to be added but not by default
* reverted some tests to use old api to prove it still works and reduce amount of change

PR #138, opened 2022-06-21
feature: basic @oneOf generation support

bzbetty:
If an input type is decorated with [GraphQLOneOfAttribute] it injects the @oneOf directive.

Not sure if the server side needs any more than that at present, could add in some extra validation around forcing all fields to be nullable and potentially that incoming variables obey the oneOf directive?

@oneOf spec - https://github.com/graphql/graphql-spec/pull/825


lukemurray:
Nice, this one I have had a need for :) 

So this is just outputting the schema right? Not yet handling server side the potential of the argument being one of many different types?

I think when we pick a type to map the input to we need to have a list of possible types and find the right type. Also what will this look like in C# for a mutation or field argument?

bzbetty:
So the interesting thing about @oneOf (from what I can tell anyway) is that the server does very little.  It's not a union type it's just a bunch of properties on one type that semantically aren't allowed to have multiple with values.  Which means you could effectively already do something similar with a bit of discipline (Input Type where you only ever set one property to have a value).

So we could add in validation for the schema to ensure all props are nullable and validation on mutation call that only one is set, but we don't strictly have to before we gain a lot of value.

Clients / code gen pay more attention to it as they can generate types that prevent multiple values being set.

eg

```
input MutationArgs @oneOf {
  Arg1: String
  Arg2: Int
}
```

becomes the following typescript

```
export type MutationArgs =
  { Arg1: Scalars['String']; Arg2?: never; }
  |  { Arg1?: never; Arg2: Scalars['Int']; };
```

but the mutation method would still end up like

```
public void Mutation(MutationArgs args) {
  if(args.Arg1.hasValue) {
  } else {
    //do something with args.Arg2
  }
}
```

lukemurray:
@bzbetty you might be best to resolve the conflicts here. Let me know if you need a hand

bzbetty:
updated with latest from master

lukemurray:
@bzbetty have you had a chance to test this with mutations using `[MutationArguments]` and mutations that just use inline/top level parameters in the method arguments?

bzbetty:
it only impacts input types so should have no effect on MutationArguments/top level params.

PR #141, opened 2022-06-23
Multiple inheritance and code cleanup

bzbetty:
Includes changes from https://github.com/EntityGraphQL/EntityGraphQL/pull/139/

* Cleanup SchemaType to use an enum instead of lots of boolean type variables (controversial? i tried to not break any public api)
* Allow types to inherit from multiple base classes/interfaces
* Cleanup Interfaces api a bit - added a "AddAllBaseTypes, AddBaseType<T> and AddBaseType(string)" which provides a lot more flexiblity (there were issues when a type was auto detected before you could add its base type)


bzbetty:
Also updated docs

Issue #146, opened 2022-07-02
Allow GraphQL mutation arguments as seperate arguments in the method signature


lukemurray:
Making the arguments on the mutation method the top level arguments in the GraphQL schema instead of having to wrap all the arguments in classes.

PR #151, opened 2022-07-05
feature: Allow GraphQL mutation arguments as seperate arguments

bzbetty:
#146

Although I feel like it could have broken something.
```
 [GraphQLMutation]
public Person AddPersonSeparateArguments(string name, List<string> names, InputObject nameInput, Gender? gender)
{
  return new Person { Name = string.IsNullOrEmpty(name) ? "Default" : name, Id = 555, Projects = new List<Project>() };
}
```

becomes
```
addPersonSeparateArguments(name: String, names: [String!], nameInput: InputObject, gender: Gender): Person
```

and

```
[GraphQLMutation]
public Person AddPersonSingleArgument(InputObject nameInput)
{
  return new Person { Name = string.IsNullOrEmpty(nameInput.Name) ? "Default" : nameInput.Name, Id = 555, Projects = new List<Project>() };
}
```

becomes

```
addPersonSingleArgument(nameInput: InputObject): Person
```

both with auto add input types on.

### DIFF
--- a/src/EntityGraphQL/Compiler/Util/ArgumentUtil.cs
+++ b/src/EntityGraphQL/Compiler/Util/ArgumentUtil.cs
@@ -103,7 +103,7 @@ public static class ArgumentUtil
         return argumentValues;
     }
 
-    private static object? BuildArgumentFromMember(ISchemaProvider schema, IReadOnlyDictionary<string, object>? args, string memberName, Type memberType, object? defaultValue, IList<string> validationErrors)
+    public static object? BuildArgumentFromMember(ISchemaProvider schema, IReadOnlyDictionary<string, object>? args, string memberName, Type memberType, object? defaultValue, IList<string> validationErrors)
     {
         string argName = memberName;
         // check we have required arguments
--- a/src/EntityGraphQL/Schema/ArgType.cs
+++ b/src/EntityGraphQL/Schema/ArgType.cs
@@ -45,6 +45,13 @@ namespace EntityGraphQL.Schema
             return arg;
         }
 
+        public static ArgType FromParameter(ISchemaProvider schema, ParameterInfo prop, object? defaultValue, Func<string, string> fieldNamer)
+        {
+            var arg = MakeArgType(schema, prop.Member, prop.ParameterType, prop.Member, defaultValue, fieldNamer);
+
+            return arg;
+        }
+
         public static ArgType FromField(ISchemaProvider schema, FieldInfo field, object? defaultValue, Func<string, string> fieldNamer)
         {
             var arg = MakeArgType(schema, field, field.FieldType, field, defaultValue, fieldNamer);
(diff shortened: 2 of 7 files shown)


bzbetty:
The second example felt off to me, I mean i like it i really want top level argument Input types (would make my graphql much easier to write), but it surprised me a bit that it turned out that way.

bzbetty:
Wondering if this feature should be opt in.

Either by a new argument to

AddMutationsFrom(autoAddInputTypes: false, addMethodsWithoutMutationAttribute: false, argumentBinding: bindingtype.MutationArgument | bindingTyping.Direct)

Or via the mutation attribute (less preferable as I don't use it now)

[GraphQLMutation(bindingtype.MutationArgument | bindingTyping.Direct)]

Thoughts?

Also trying to think of what tests still need adding

* behaviour with autoAddInputTypes = false
* Behaviour when not in class with other mutations (eg one of the other mutations in the test class may have bound something for me)
* does it still work with method level injection/dbcontext


Also feel like @breyed may want to input on this one since it came from his original ticket.

breyed:
I think the single-parameter example is fine. There will be use cases where initially a mutation only has a single parameter of object type, but as the API evolves, other parameters will be added.

bzbetty:
Also question for @lukemurray:
* Currently I'm grabbing the values directly from docVariables instead of gqlRequestArgs/ ArgumentUtil.BuildArgumentsObject as it was easier at the time.  However I'm assuming the other way is needed to support some usecase that I'm not aware of, any reason the current implementation did it that way?

lukemurray:
Looks good!

`docVariables` only include variables pass in the query document e.g. `$thisVar` below

```
query MyQuery($thisVar: ID!) {...}
```

Whereas each field may have other defined arguments, either set inline in the query like `sortByName` below
```
query MyQuery($thisVar: ID!) {
   person(id: $myVar sortByName: true) {...}
}
```

There could also be an optional argument on that field that has a default value.

Hence if we only take `docVariables` you have not resolved all the arguments and may have some missing for a valid query.

Each of those bits are "known" at different times, hence it is "resolved" when the field expression is being built.

bzbetty:
**Added Test**
* behaviour with autoAddInputTypes = false
* only primitives
* inline values in gql

**To Test Still**
* Behaviour when not in class with other mutations (eg one of the other mutations in the test class may have bound something for me)
* does it still work with method level injection/dbcontext (do we need some attributes for injected items, or do we disable injection all together at this level?)
* still works with validators
* async + IQueryable
* with default argument values

lukemurray:
@bzbetty Good question.

Should they? Yes I feel they should. Assuming nested works in asp.net validation to?

Does it currently work in nested objects in 2.x.0 ? I'm not actually sure! Need some more tests there :)

bzbetty:
> Does it currently work in nested objects in 2.x.0 ? I'm not actually sure! Need some more tests there :)

Does not appear to work currently, validation attributes are detected for top level ArgTypes only and there's no where to store them for child objects or lists.

> Should they? Yes I feel they should. Assuming nested works in asp.net validation to?

Yeah it works in MVC, but it has a whole visitor pattern setup to spider the object which would be a lot to replicate.  Maybe people need to resort to custom validation at that point.

I do feel like the attribute based one could be done a bit better/flexible by storing any ValidationAttribute in a list and looping over it, although there is a special case for required attribute due to the IsRequired property too.

PR #153, opened 2022-07-06
mutation controllers always created per request

lukemurray:
Instead of looking a mutation controller up in the service provider which can lead to mis-matched lifetimes of services. Always create a mutation controller per request (using DI for constructor) like asp.net controllers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Breaking changes:
 - Removed obsolete `ISchemaType.BaseType`. Use `ISchemaType.BaseTypes`
 - Cleaned up `SchemaType` constructors - using `GqlTypeEnum` instead of many boolean flags
 - Removed obsolete `SchemaProvider.AddInheritedType<TBaseType>`
+- Removed the instance parameter from `AddMutationsFrom` and friends. Mutation "controllers" are now always created per request like an asp.net controller. Use DI for any constructor parameters
 
 Changes:
 
--- a/docs/content/schema-creation/01-mutations.md
+++ b/docs/content/schema-creation/01-mutations.md
@@ -46,9 +46,9 @@ You can add the mutation controller to a schema in the following ways:
 schema.AddMutationsFrom<PeopleMutations>();
 ```
 
-EntityGraphQL adds the `PeopleMutations` mutation controller and all its mutation methods (those with `[GraphQLMutation]` applied).
+EntityGraphQL adds the `PeopleMutations` mutation controller and all its mutation methods (those with `[GraphQLMutation]` applied) to the schema.
 
-For each mutation request, EntityGraphQL creates a new instance of `PeopleMuations` and provides services to it through [dependency injection](#dependenciesinjection&services).
+For each mutation request, EntityGraphQL creates a new instance of `PeopleMuations` and provides services to its constructor through [dependency injection](#dependenciesinjection&services).
 
 **Register all mutation controllers implementing or derving from a type**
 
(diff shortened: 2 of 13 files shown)


lukemurray:
@breyed / @bzbetty  I think this cleans things up and makes sure the mismatched lifetime of services doesn't happen. Make them much similar to controllers.

Thoughts?

lukemurray:
Yes I'm fine with including `Microsoft.Extensions.DependencyInjection` packages as dips for the core library. It provides DI helpers like this.

breyed:
The [commit with 01-mutations.md](https://github.com/EntityGraphQL/EntityGraphQL/pull/153/commits/f3a87f7bfd0603f0bddade932732f0dace88866b#diff-90d77e968f9a8dd2a2c9d77899d7458cfd4b93d042072c9237db3f04e2153e5b) has:
> "provides services **to its constructor** through dependency injection

Does this mean we're dropping the option to provide services through method parameters? I don't think there's any need to do so. At startup, you can walk through each parameter of a mutation method and check whether its type is that of a registered service. All mutation method parameters must be either input types or services (but not both).

``` c#
var servicesByType = services.ToDictionary(s => s.ServiceType); // Used across all mutations.

foreach (var parameter in methodParameters) {
   bool isService = servicesByType.ContainsKey(parameter.Type);
   bool isGraphQLMutationParameter =
      graphQLInputTypesByClrType.ContainsKey(parameter.Type) || 
      autoAddInputTypes && !isService;
   if (isService && isGraphQLMutationParameter) throw new InvalidOperationException("A mutation method parameter cannot be both a dependency injection service and a GraphQL input type.");
   // … Record services and parameters for use at request time …
}
```

lukemurray:
@breyed No, I will make the clearer. It is purely saying _if_ you need to have parameters in the constructor of the mutation class, it is via DI. It still does DI for the methods too.

breyed:
Using one mutation controller per request without the mutation controller needing to be a DI service will be a nice simplification. It may resolve a bug, too. Yesterday, I ran into an error when trying to make a mutation controller that took a singleton service in its constructor:

``` c#
class MyMutation {
   public MyMutation(IMySingletonService service) {}
   // …
}
```

When I registered the mutation controller as a transient:
``` c#
services.AddTransient<MyMutation>();
```

I got an error at startup. I had someone waiting on an update, so I backed out the change and went back to the old way of manually creating the mutation controller instance and passing services as mutation method parameters. I forgot to record what the error was, if with the change to a mutation controller not being a service, the problem will disappear anyway.

lukemurray:
The issue I think you hit will be gone. As `MyMutation` will be create per request with this change and each time `IMySingletonService` will be injected from the service provider.

You no longer have to do `services.AddTransient<MyMutation>();`

Just add the mutation type to the schema and register `IMySingletonService`

breyed:
Do we need `ActivatorUtilities`? Are there any cases where EntityGraphQL creates instances of services itself? I would expect EntityGraphQL to call `ServiceProvider.GetRequiredService`. The service provider creates the service as needed (e.g. if it's transient) and disposes it at the appropriate time in the pipeline. All EntityGraphQL needs to do is get services, never activate them. Or am I missing something?

lukemurray:
Per request, we need to create an instance of the mutation class/controller. That may have parameters in the constructor which need to be resolved. `ActivatorUtilities` is a helper method that uses the service provider to resolve the parameters for creating the class.

E.g. it will use ServiceProvider to create a new instance of the parameter if needed. It will be whatever you have registered in services.

We could remove the dependancy and write code that looks at all parameters in the constructor(s) and searches the service provider for them and them calls the correct Activator.CreateInstance. It is similar to what we do in mutation methods, but different as we have other things to look for and other logic. I think this dependency is small and built for this use case.

breyed:
I'm on board with activating the service, but I'm questioning whose responsibility it is to do the activation. My understand is that it's `ServiceProvider`'s responsibility, since it understands the scope. For example, for scoped-lifetime services, it knows when the web request begins and end. It also knows if it's already activated an instance of the service for that scope. In the case of a transient service, it will always activate a new instance, but it's `ServiceProvider` that looks up the service lifetime and performs the activation logic, not EntityGraphQL.

lukemurray:
Are you looking at this line?

`ActivatorUtilities.CreateInstance(serviceProvider, method.DeclaringType)`

As that is what happens. 

`method.DeclaringType` is `MyMutation` in the above example. We do not want `ServiceProvider` to resolve that, as it introduces the lifetime issues you highlighted, similar to Controllers.

`ActivatorUtilities.CreateInstance` creates an instance of type `method.DeclaringType` (e.g. `MyMutation`) using `ServiceProvider` to resolve any parameters needed to create that class. See here https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance?view=dotnet-plat-ext-6.0#microsoft-extensions-dependencyinjection-activatorutilities-createinstance(system-iserviceprovider-system-type-system-object())

breyed:
Got it. We're using `ActivatorUtilities.CreateInstance` to create the mutation controller itself. I was mistakenly thinking we were using it to create the injected services. Looks good.

PR #156, opened 2022-07-07
#155 clean up interfaces

### LINKED ISSUE #155: Clean up GraphQL schema Interfaces
Make sure we follow the spec

- [x] You can only implement interfaces not types
- [x] Allow multiple implements via the & operator on a type or interface

Rename methods to align with graphql language

- Keep schema.AddInterface<Character>() to directly add interfaces to the schema

- [x] Add SchemaType.Implements<Character>(bool addIfNotInSchema = true, bool addAllFieldsIfAddedToSchema = true) to allow you tell the type any interfaces it implements. I.e. schema.AddInterface<Character>().Implements<T>() or schema.AddType<Human>().Implements<Character>()
- [x] Rename SchemaType.AddAllBaseTypes to ImplementAllBaseTypes(bool autoAddAllFieldsInInterface = true) - Adds the base classes & interfaces of the dotnet type to the schema if not there and makes the schema type implement them
- [x] Remove SchemaType.AddBaseType<>() and use .Implements()

lukemurray:
Closes #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Breaking changes:
 - Removed the instance parameter from `AddMutationsFrom` and friends. Mutation "controllers" are now always created per request like an asp.net controller. Use DI for any constructor parameters
 - Renamed `ISchemaType.AddBaseType` to `ISchemaType.Implements` to align with GraphQL language
   - `ISchemaType.Implements` will throw an exception if you try to implement a non-interface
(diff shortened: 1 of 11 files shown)


lukemurray:
@bzbetty I think this covers it all? Ended up just being renaming and added the parameters to allow auto adding types/fields or not.

I think you asked if there should be an option to auto do this in the SchemaBuilder?

I think that makes sense.

bzbetty:
Yes looks good to me, and yes auto option in Schema Builder

PR #158, opened 2022-07-07
SchemaBuilder options

### LINKED ISSUE #157: Add option in SchemaBuilder to auto add and implement interfaces
As title. Add the option in SchemaBuilder to auto create all the interfaces. 

Look at extracting all the options into an options class as the parameters keep growing.

lukemurray:
Close #157 

- [x] Move `SchemaBuilder` option parameters to an options class as they are growing long
- [x] Add new option to create interfaces etc during the reflection of `SchemaBuilder.FromObject`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Breaking changes:
 - Renamed `ISchemaType.AddBaseType` to `ISchemaType.Implements` to align with GraphQL language
   - `ISchemaType.Implements` will throw an exception if you try to implement a non-interface
 - Renamed `ISchemaType.AddAllBaseTypes` to `ISchemaType.ImplementAllBaseTypes` to align with GraphQL language
+- `Create` & `FromObject` on `SchemaBuilder` now take option classes to configure the create of the schema through reflection
+  - `ISchemaType.AddAllFields` also takes the option class to configure it's behaviour
+  - `ISchemaType.AddAllFields` default behaviour now auto adds any complex types found really reflection the properties & fields and will add those to the schema
 
 Changes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Breaking changes:
 - `Create` & `FromObject` on `SchemaBuilder` now take option classes to configure the create of the schema through reflection
   - `ISchemaType.AddAllFields` also takes the option class to configure it's behaviour
   - `ISchemaType.AddAllFields` default behaviour now auto adds any complex types found really reflection the properties & fields and will add those to the schema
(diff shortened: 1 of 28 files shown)


bzbetty:
seems like a nice direction to prevent API breakages

lukemurray:
@bzbetty keen to get your thoughts on this. Does it follow what you would have thought?

note there are 2 things (1 existed)

`AutoCreateNewComplexTypes` will also create interfaces if it finds a field that returns an interface.

`AutoCreateInterfaceTypes` is purely for adding base types/ interfaces of a type (regardless of it being used in a field return type).

bzbetty:
Mostly like it.

* Should AutoCreateInterfaceTypes be split? it seems to both create interface types and implement them? (AutoImplementAllBaseTypes?)
* should the parameters that are passed to ImplementAllBaseTypes come from options too? (addTypeIfNotInSchema, addAllFieldsOnAddedType)


Probably not for this PR, but starting to wonder if we want some hooks to provide a bit of customizability to the auto builder.  

examples
* instead of just boolean options maybe pass in Func<x, bool> so people could ignore specific items? (maybe interfaces just need to support for GraphQLIgnore)
* In a similar vein TypeAdded/FieldAdded hooks - SchemaBuilderOptions.OnFieldAdded = (Field) => {}
* a way to apply something to all fields/types? (eg I wanted to add auth to every field the other day)

lukemurray:
Maybe `AutoCreateInterfaceTypes` should be named `AutoImplementBaseTypes`?

Thinking it through. If we have `AutoCreateNewComplexTypes`, `AutoCreateInterfaceTypes` & `AutoImplementAllBaseTypes`

if `AutoImplementAllBaseTypes` is `true` but `AutoCreateInterfaceTypes` is false you'll hit errors as it tries to implement types that do not exist?

`AutoCreateNewComplexTypes` currently "creates" (adds type and it's fields) for a field return type. This could be a class or an interface/abstract class.

I do like the idea of the callbacks. There is a list of field name to ignore, callback provides more flexibility

lukemurray:
So you want it to add the type (and fields?) of derived typed but not implement them?

bzbetty:
As in
```
public class Context {
   IDbSet<Product> Products {get; set;}
}

public abstract class Product {
}

public class Book : Product {}
public class Movie : Product {}
```

currently after doing a SchemeBuilder.FromObject<Context>() I then have to

```
schema.AddType<Book>().ImplementAllBaseTypes().AddAllFields();
schema.AddType<Movie>().ImplementAllBaseTypes().AddAllFields();
```

Was wondering if it was worth automatically finding the subclasses and adding them.  That's obviously harder than adding base types as it invovles assembly scanning which might not be worth it (although we do that when registering mutations at present).

it's a minor thing, I've just found a couple of times I've either missed it, or had to remove the line if it was ever referenced in the context later.

lukemurray:
Ahh interesting. If we did find them I guess you run the risk of finding many others that you don't intend. But if it is documented and understood I could see it being helpful.

lukemurray:
I'll leave that to a separate PR though.

PR #160, opened 2022-07-10
feature: Data annotations validator: validating nested objects

bzbetty:
Had a bit of a go at making validation of nested objects and lists working.

* Implemented IArgumentValidator with a basic class that does recursive validation
* Auto register this validator type with BaseField
* should cover all validation attribute types not just the ones you had before

I played around with removing the validation attributes from ArgTypes since it should cover the same things however

* couldn't replace required field validation on non nullable fields as it didn't have access to raw arguments (eg it was getting an object with default values which are technically valid), could probably remove the range/length attribute checking from argtype though just means they'd be caught later
* couldn't find a nice way to remove the throw statements from ArgumentUtil as it would allow all fields to be validated instead of bailing early for certain types of errors (eg required fields always checked first as per above).

### DIFF
--- a/src/EntityGraphQL/Compiler/EntityGraphQLValidationException.cs
+++ b/src/EntityGraphQL/Compiler/EntityGraphQLValidationException.cs
@@ -10,7 +10,7 @@ public class EntityGraphQLValidationException : Exception
 
     public EntityGraphQLValidationException(IEnumerable<string> validationErrors)
     {
-        ValidationErrors = validationErrors.ToList();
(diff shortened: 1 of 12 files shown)


bzbetty:
updated argumentutil to pass validation errors out instead of just throwing, that way we can combine them later

bzbetty:
extended to support:
*  top level arrays (eg the allArgs object)
* validating the object against the method info (validation attributes on parameters)

lukemurray:
@bzbetty , sorry, could you resolve the conflicts please? Looking to merge this one next :)

PR #174, opened 2022-07-20
PoC: Summarize Field Extension

bzbetty:
first take at a new field extension for providing summary/aggregate data.

```
query() {
      people() { 
          summarize { average { height } }
      }
  }
```

**Supports:**
* Min, Max, Average, Sum, Count
* Number fields only

**Could Potentially Support**
* Grouping before doing stats? (haven't really thought that one through yet)
* Filtering before doing stats
* Stats on other types (eg min/max string?)

**Notes**
* Naming very much open to change - eg could be aggregate or stats etc
* Tried adding minimal extra types 
** adds a field to existing type (eg people gets a summarize field which may not be the best)
** reuses the same type to put the stats on (mainly a shortcut, but isn't great when it means theres fields you can't query)

### DIFF
--- a/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
+++ b/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
@@ -521,7 +521,7 @@ namespace EntityGraphQL.Compiler.Util
             return mi;
         }
 
-        private static Expression CreateNewExpression(string fieldDescription, Dictionary<string, Expression> fieldExpressions)
+        public static Expression CreateNewExpression(string fieldDescription, Dictionary<string, Expression> fieldExpressions)
         {
             var fieldExpressionsByName = new Dictionary<string, Expression>();
             foreach (var item in fieldExpressions)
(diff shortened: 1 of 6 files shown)


lukemurray:
Cool. Did you hit any issue doing this with field extensions? Including lack of documentation etc?

I don't mind `summarize` but I think I prefer `aggregate`?

bzbetty:
> Cool. Did you hit any issue doing this with field extensions? Including lack of documentation etc?

Documentation was mostly go look at the examples if I remember rightly, that was all pretty easy.  Figuring out the expression trees was a pain though (although that's pretty much always the case with expression trees).

> I don't mind `summarize` but I think I prefer `aggregate`?

yeah i was kinda leaning the same way, i think other gql tend to use aggregate too.

Also starting to think it should be a top level field instead of a subfield, too hard to work out how it would interact with everythign else (pagination/filtering etc) at which point I'm not sure if its really a field extension or something new.

Still pondering this one a bit, not really in a rush.

PR #177, opened 2022-07-21
feature: union types, and a few fixes for interfaces

bzbetty:
Feel like some of the API is getting a bit swamped now - eg Schema type now has a AddPossibleType method which is only valid on interfaces or unions.  Wondering if need to think about splitting types?


lukemurray:
That is a fair point regarding the Schema type class. Think I'll try to wrap up a 3.x release with the current breaking changes first as there is a bit in there now. These can come with 3.1/2/etc and give time to think and plan out breaking changes (if needed) for 4.x.

bzbetty:
hmm union types seem to break the assumption that field names are unique and of the same type.

lukemurray:
What is the hack in `VisitInlineFragment`?

I haven't read the spec in a while on union types, does it same what happens with duplicate field names? Do you have to use the inline fragment syntax to tell it which type to select the field from?

bzbetty:
The hack in VisitInlineFragment was to clone the context object with a different fragment parameter then copy the fields back onto the original one.

Essentially we end up with one GraphQLNode with fields on it for all the types, then when putting the expression back together I split it back out by looking at the field parameter type.

I'm guessing it would be better to keep it separate by using a new QueryField type that supports fragments, but would still likely need the dictionary removed at some point.  I haven't dove into it to see how far the dictionary gets passed around, maybe it wasnt' as deep as I thought.

Unions don't have fields on them themselves (except you can query __typename).  So pretty much everything has to be asked for in fragments, none of the fields need to be shared.  I'm not 100% sure if it's valid that two fields with the same name have different types, but at the very least they're going to have rather different expressions.

bzbetty:
It appears the fields can all be different types, for example:

```
union Character = Lysander | Hermia | Demetrius | Helena

type Lysander  {
   loves: Hermia!
}

type Hermia {
  loves: Lysander!
}

type Helena {
  loves: Demetrius!
}

type Demetrius {
  loves: Hermia!
}
```

bzbetty:
* Implemented a GraphQLInlineFragmentField and used the Expand method to correctly handle inline fragments.

Then i tried two experiments:

* Replace the Dictionary<string, CompiledField> with an IEnumerable<CompiledField> it mostly worked but there were a couple of odd test failures and I wasn't happy with the API changes to field extensions

* Kept the Dictionary but prefixed field names with their types (eg instead of the key being 'name' it became 'Dog.name'). Which i then removed later when creating the final type.  Bit of a hack, but a much smaller change and nicer to work with.

I prefer the second one, happy for others to weigh in (including other implementation ideas)


There's currently a small issue with one of the tests seeming to be a bit flaky depending on the order they're run, guessing it's caching in the dynamic type builder and that I need an extra parameter to ensure types are unique.

docs/content/schema-creation/03-types.md
```diff
@@ -161,6 +161,34 @@ query {
...
+    ...
+}
+
+public class Droid : ICharacter {
+    ...
+}
+
+
+// creating our schema
+var union = schema.AddUnion<Character>(name: "Character", description: "represents any character in the Star Wars trilogy");
```
lukemurray, line 186: `Character` here I assume is `ICharacter` ?

src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
```diff
@@ -374,22 +374,11 @@ protected override void VisitInlineFragment(InlineFragmentNode node, IGraphQLNod
                 if (type != null)
                 {
                     var fragParameter = Expression.Parameter(type.TypeDotnet, $"frag_{type.Name}");
-                    IGraphQLNode newContext = context switch
-                    {
-                        GraphQLListSelectionField graphQLListSelectionField => new GraphQLListSelectionField(graphQLListSelectionField!, fragParameter),
```
lukemurray, line 379: Wasn't this added to support the interfaces?

Reading it again I assume it was not correctly expanding fragments in interfaces?

src/EntityGraphQL/Compiler/GqlNodes/GraphQLInlineFragmentField.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class GraphQLInlineFragmentField : BaseGraphQLField
+    {
+        public GraphQLInlineFragmentField(ISchemaProvider schema, string name, Expression? nodeExpression, ParameterExpression rootParameter, IGraphQLNode parentNode)
+            : base(schema, null, name, nodeExpression, rootParameter, parentNode, null)
+        {
+        }
+
+        public override void AddField(BaseGraphQLField field)
+        {
+            field.Name = $"{Name}.{field.Name}";
```
lukemurray, line 19: Why is this needed?

lukemurray:
This is looking good. Thanks.

Need to resolve conflicts. I made a couple of comments. Nothing major.

src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
bzbetty: it was, but at that point I hadn't understood how the expand function was working for other fragments.  The new way is a lot more inline with how you were originally doing it and felt much less like a hack than what I had going.

src/EntityGraphQL/Compiler/GqlNodes/GraphQLInlineFragmentField.cs
bzbetty: This is because multiple 'subtypes' might have a field of the same name but different expressions, so this prevents them overwriting each other (it is removed later during actual selection).  Could have potentially merged the expressions instead using a IS check, but was difficult to get it working at that spot in the code.

lukemurray: Can you point me to where this is undone?

src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
```diff
@@ -450,27 +451,24 @@ public static Expression MakeSelectWithDynamicType(string fieldDescription, Para
...
+                Expression? previous = Expression.Constant(null, baseDynamicType);
                 foreach (var type in validTypes)
                 {
 
-                    var fieldsOnType = fieldExpressions
-                       .Where(i => RootType(i.Value)!.IsAssignableFrom(type) || RootType(i.Value) == typeof(ISchemaType))
-                       .ToDictionary(i => i.Key, i => i.Value);
+                    var fieldsOnType = fieldExpressions.Values
+                       .Where(i => RootType(i.Expression)!.IsAssignableFrom(type) || typeof(ISchemaType).IsAssignableFrom(RootType(i.Expression)))
+                       .ToDictionary(i => i.Field.Name.Replace($"{RootType(i.Expression)?.Name}.", ""), i => i.Expression);
```
bzbetty, line 468: undone here

src/EntityGraphQL/Compiler/GqlNodes/GraphQLInlineFragmentField.cs
bzbetty: added a comment to the line that undoes it

Issue #182, opened 2022-07-31
No schema type found for dotnet type Char. Make sure you add it or add a type mapping


bzbetty:
Upon upgrading to 3.0 I get this error while deep in EQL before I can manually add the type.

```
   at EntityGraphQL.Schema.SchemaProvider`1.GetSchemaType(Type dotnetType, QueryRequestContext requestContext)
   at EntityGraphQL.Schema.SchemaBuilder.CacheType(Type propType, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.<ProcessFieldOrProperty>d__6.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at EntityGraphQL.Schema.SchemaBuilder.GetFieldsFromObject(Type type, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.CacheType(Type propType, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.<ProcessFieldOrProperty>d__6.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at EntityGraphQL.Schema.SchemaBuilder.GetFieldsFromObject(Type type, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.CacheType(Type propType, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.<ProcessFieldOrProperty>d__6.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at EntityGraphQL.Schema.SchemaBuilder.GetFieldsFromObject(Type type, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.CacheType(Type propType, ISchemaProvider schema, SchemaBuilderOptions options, Boolean isInputType)
   at EntityGraphQL.Schema.SchemaBuilder.<ProcessFieldOrProperty>d__6.MoveNext()
[...]

PR #186, opened 2022-08-01
Replicate bug #182

bzbetty:
simple change to replicate #182 - this small change makes about half the tests in the project fail with no obvious fix

seems easy enough to fix, just wasn't sure if we want to auto register scalar types for primitives, ignore the type, or do something else.

### DIFF
--- a/src/tests/EntityGraphQL.Tests/TestDataContext.cs
+++ b/src/tests/EntityGraphQL.Tests/TestDataContext.cs
@@ -90,6 +90,7 @@ namespace EntityGraphQL.Tests
     public class Project
     {
         public int Id { get; set; }
+        public char Code { get; set; }
         public string Name { get; set; }
         public int Type { get; set; }
         public Location Location { get; set; }


lukemurray:
Ahh interesting. Or map it to another type.

This is the default scalars and mappings

```
            // default GQL scalar types
            schemaTypes.Add("Int", new SchemaType<int>(this, "Int", "Int scalar", null, GqlTypeEnum.Scalar));
            schemaTypes.Add("Float", new SchemaType<double>(this, "Float", "Float scalar", null, GqlTypeEnum.Scalar));
            schemaTypes.Add("Boolean", new SchemaType<bool>(this, "Boolean", "Boolean scalar", null, GqlTypeEnum.Scalar));
            schemaTypes.Add("String", new SchemaType<string>(this, "String", "String scalar", null, GqlTypeEnum.Scalar));
            schemaTypes.Add("ID", new SchemaType<Guid>(this, "ID", "ID scalar", null, GqlTypeEnum.Scalar));

            // default custom scalar for DateTime
            schemaTypes.Add("Date", new SchemaType<DateTime>(this, "Date", "Date with time scalar", null, GqlTypeEnum.Scalar));

            customTypeMappings = new Dictionary<Type, GqlTypeInfo> {
                {typeof(sbyte), new GqlTypeInfo(() => Type("Int"), typeof(sbyte))},
                {typeof(short), new GqlTypeInfo(() => Type("Int"), typeof(short))},
                {typeof(ushort), new GqlTypeInfo(() => Type("Int"), typeof(ushort))},
                {typeof(long), new GqlTypeInfo(() => Type("Int"), typeof(long))},
                {typeof(ulong), new GqlTypeInfo(() => Type("Int"), typeof(ulong))},
                {typeof(byte), new GqlTypeInfo(() => Type("Int"), typeof(byte))},
                {typeof(uint), new GqlTypeInfo(() => Type("Int"), typeof(uint))},
                {typeof(float), new GqlTypeInfo(() => Type("Float"), typeof(float))},
                {typeof(decimal), new GqlTypeInfo(() => Type("Float"), typeof(decimal))},
                {typeof(byte[]), new GqlTypeInfo(() => Type("String"), typeof(byte[]))},
                {typeof(bool), new GqlTypeInfo(() => Type("Boolean"), typeof(bool))},
            };
```

Should `char` just be mapped to another schema type by default?

bzbetty:
I think mapping it to its own scalar gives the most flexibility, that way the client can decide how to handle it.

Based on that list it looks to be about the only primitive not covered.

Curiously enough it didn't explode in 2.3

lukemurray:
Thanks for this. I have added it and the new default scalar

PR #189, opened 2022-08-02
Was using default enum value for fully qualified enums instead of the parsed value

bzbetty:
oops.

### DIFF
--- a/src/EntityGraphQL/Compiler/EntityQuery/EntityQueryNodeVisitor.cs
+++ b/src/EntityGraphQL/Compiler/EntityQuery/EntityQueryNodeVisitor.cs
@@ -140,7 +140,6 @@ namespace EntityGraphQL.Compiler.EntityQuery
             if (!schemaType.HasField(field, requestContext))
             {
                 var enumOrConstantValue = constantVisitor.Visit(context);
-
                 if (enumOrConstantValue == null)
                 {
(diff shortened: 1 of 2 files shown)


src/EntityGraphQL/Compiler/EntityQuery/EntityQueryNodeVisitor.cs
```diff
@@ -149,12 +148,17 @@ public override Expression VisitIdentity(EntityQLParser.IdentityContext context)
...
                 }
 
                 return enumOrConstantValue;
             }
+
+            if(schemaType.IsEnum)
+            {
+                return Expression.Constant(Enum.Parse(schemaType.TypeDotnet, context.GetText()));                
+            }
+
```
lukemurray, line 161: The `ConstantVisitor` is trying to handle this.

https://github.com/EntityGraphQL/EntityGraphQL/blob/master/src/EntityGraphQL/Compiler/ConstantVisitor.cs#L74

I wonder why it is returning `null`?

bzbetty: It skips the block the second time through due to the schema.hasfield check.

I'd get the constant visitor to do this but it doesn't have the schema type passed to it, so seemed easier inline.

Only covers the case of a fully qualified enum.

PR #193, opened 2022-08-04
Schema Directives refactoring

bzbetty:
Starting to refactor towards handling new schema directives (currently theres @deprecated and @oneOf).

Ultimately I was hoping to be able to include @auth and @required directives in my SDL (given a bunch of tools support it) but it's not an official standard.

~~A breaking change since I remove a couple of setters (although that could probably be reverted)~~

Currently trying to figure out if this a valuable refactor given that graphql doesn't currently support custom schema directives (lots of discussions on it from what I see but nothing final)

Thinking it might be nice to allow extensibility in the SchemaGenerator.

### DIFF
--- a/docs/docs/directives/schema-directives.md
+++ b/docs/docs/directives/schema-directives.md
@@ -11,8 +11,8 @@ Schema Directives are used to decorate the schema to provide more information to
 The GraphQL spec defines directives that are supported out of the box in EntityGraphQL.
 
 - `@deprecated(reason: String)` - Tells the client that this type, field or enum value should no longer be used along with the reason why or a suggested alternative
-- `@oneOf` - Skip this field if the argument is true.
-- `@specifiedBy(url: String)` - Skip this field if the argument is true.
+- `@oneOf` - Mark the input type as a type where only one of it's fields should ever be non-null
+- `@specifiedBy(url: String)` - Used to provide a scalar specification URL for specifying the behavior of custom scalar types.
 
 ## Deprecated
 
@@ -50,7 +50,7 @@ Although each field on the input is nullable the `@oneOf` input type has one fur
 
 ## Specified By
 
-The [@specifiedBy](https://spec.graphql.org/draft/#sec--specifiedBy) directive is used to provide extra information about a custom scalar type.
+The [@specifiedBy](https://spec.graphql.org/draft/#sec--specifiedBy) directive is used to provide extra information about a custom scalar type. The URL should point to a human-readable specification of the data format, serialization, and coercion rules. It must not appear on built-in scalar types.
 
 ## Custom Directives
 
--- a/docs/docs/directives/schema-directives.md
+++ b/docs/docs/directives/schema-directives.md
@@ -1,0 +1,133 @@
+---
+sidebar_position: 1
+---
+
+# Schema Directives
+
+Schema Directives are used to decorate the schema to provide more information to clients, EntityGraphQL can also use this information to update the introspection model or provide extra functionality such as validation.
+
+## Built-In Directives
+
+The GraphQL spec defines directives that are supported out of the box in EntityGraphQL.
+
+- `@deprecated(reason: String)` - Tells the client that this type, field or enum value should no longer be used along with the reason why or a suggested alternative
+- `@oneOf` - Skip this field if the argument is true.
+- `@specifiedBy(url: String)` - Skip this field if the argument is true.
+
+## Deprecated
+
+Fields, Enum Values, Mutation Arguments and Types marked with the c# [ObsoleteAttribute] will have the [@deprecated](https://spec.graphql.org/draft/#sec--deprecated) directive automatically added.
+
+You can also use the Deprecate(string reason) extension method on a IField.
+
+## One Of Input Types
+
+EntityGraphQL supports [One Of Input Types](https://github.com/graphql/graphql-spec/pull/825).
+
+Mark an input type with `GraphQLOneOfAttribute` and EntityGraphQL will mark the type with `@oneOf` in the schema and validate the input meets the requiements on execution.
+
+```cs
+[GraphQLOneOf]
+private class OneOfInputType
+{
+    public int? One { get; set; }
+    public int? Two { get; set; }
+}
+```
+
+This will generate the follow graphql schema.
+
+```graphql
+input MutationArgs @oneOf {
+  one: Int
+  two: Int
+}
+```
+
+Although each field on the input is nullable the `@oneOf` input type has one further validation step.
+
+- Exactly one key must be specified
+
+## Specified By
+
+The [@specifiedBy](https://spec.graphql.org/draft/#sec--specifiedBy) directive is used to provide extra information about a custom scalar type.
+
+## Custom Directives
+
+Like [Field Extensions](../field-extensions/) you can extend the [ExtensionAttribute](../other-extensibility/extension-attribute) to apply extension to `IFields` and `ISchemaTypes` (see `GraphQLOneOfAttribute`).
+
+IFields and ISchema types also have a method `AddDirective` that is common to call in the attribute extension method or a custom extension method to register the directive.
+
+```
+public static class DeprecatedDirectiveExtensions
+{
+    public static void Deprecate(this IField field, string? reason)
+    {
+        field.AddDirective(new DeprecatedDirective(reason));
+    }
+}
+public class ObsoleteAttributeRegistration : IExtensionAttribute<ObsoleteAttribute>
+{
+    public void ApplyExtension(IField field, ObsoleteAttribute attribute)
+    {
+        field.Deprecate(attribute.Message);
+    }
+}
+```
+
+The `ISchemaDirective` method contains:
+
+* The `On` property for marking where it is valid to use this directive
+* The `ToGraphQLSchemaString()` method where you return the formatted directive including arguments (eg `return $"@deprecated(reason: \"{Reason}\")";`)
+* A series of methods the SchemaInstrospection class calls allowing you to modify the introspection result (primarily used because instrospection for @deprecated and @oneOf is done using specific fields and not yet in an extensible way)
+
+
+## Example Sample Directive
+
+Finds the ASP.NET [Authorize] attribute and adds @authorize(roles: String, policies: String) to the schema information
+
+```
+public class AuthorizeAttributeRegistration : 
+    IExtensionAttribute<AuthorizeAttribute>,
+    IExtensionAttribute<GraphQLAuthorizePolicyAttribute>
+{
+    public void ApplyExtension(IField field, AuthorizeAttribute attribute)
+    {
+        field.AddDirective(new AuthorizeDirective(attribute));
+    }
+    public void ApplyExtension(IField field, GraphQLAuthorizePolicyAttribute attribute)
+    {
+        field.AddDirective(new AuthorizeDirective(attribute));
+    }
+    public void ApplyExtension(ISchemaType type, AuthorizeAttribute attribute)
+    {
+        type.AddDirective(new AuthorizeDirective(attribute));
+    }
+    public void ApplyExtension(ISchemaType type, GraphQLAuthorizePolicyAttribute attribute)
+    {
+        type.AddDirective(new AuthorizeDirective(attribute));
+    }
+}
+public class AuthorizeDirective : ISchemaDirective
+{
+    public AuthorizeDirective(AuthorizeAttribute authorize)
+    {
+        Roles = authorize.Roles;
+        Policies = new List<string>() { authorize.Policy! };
+    }
+    public AuthorizeDirective(GraphQLAuthorizePolicyAttribute authorize)
+    {
+        Policies = authorize.Policies;
+    }
+    public IEnumerable<TypeSystemDirectiveLocation> On => new[] {                
+        TypeSystemDirectiveLocation.OBJECT,
+        TypeSystemDirectiveLocation.FIELD_DEFINITION            
+    };
+    public string? Roles { get; }
+    public List<string> Policies { get; }
+    public string ToGraphQLSchemaString()
+    {
+        return $"@authorize(roles: \"{Roles}\", policies: \"{string.Join(", ", Policies)}\")";
+    }
+}
+```
--- a/src/EntityGraphQL/Schema/BaseField.cs
+++ b/src/EntityGraphQL/Schema/BaseField.cs
@@ -6,6 +6,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using EntityGraphQL.Compiler;
 using EntityGraphQL.Compiler.Util;
+using EntityGraphQL.Schema.Directives;
 using EntityGraphQL.Schema.FieldExtensions;
 using EntityGraphQL.Schema.Validators;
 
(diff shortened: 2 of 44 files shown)


bzbetty:
https://github.com/graphql/graphql-spec/issues/300

lukemurray:
Nice. I like this.

If the spec formally adopts something slightly different the aim will be to follow that. But like `@oneOf` I'm OK adding things that does not break the spec and ands value, especially if it is already getting adopted.

bzbetty:
Realised these are quite similar to fieldextensions.

Was wondering about merging the concepts and maybe introduce type extensions at the same time?

At the very least maybe having the introspection call all the directives/extensions to let them update it.

Any thoughts?

lukemurray:
That could make sense.

Extensions are all about letting you "reuse code" to make similar changes to different fields.

In this case we are trying to make similar changes to fields - which is registering custom directives at the schema definition level to let external clients/tools know more information. So it makes sense.

If we go this path, having similar for Types also makes sense.

I guess it just needs to be simple enough, as the extensions for fields can _seem_ be complex if you're working with expressions.

Still the extensions will need a way to register in the schema the extra information. As that is how they were intended to work. Although I think you're suggesting to have the schema output code know about extensions and make calls to them to allow them to add/change information for schema output?

bzbetty:
So i really like how you registered field extensions using the custom attributes, was thinking of doing something similar with SchemaDirectives.

Problem I've hit is that a lot of the attributes I wanted to use are built in instead of custom (eg ObsoleteAttribute).

I guess i could force the user to register something like a IDirectiveProcessor (but for schema instead of operation directives) along with an optional list of valid attributes?

schemaProvider.AddDirective<DeprecatedDirective, ObsoleteAttribute>();


sidenote: Considering I think it might be possible to add a directive to the schema itself we might need to move the AddDirective method schema.Directives().Add?

lukemurray:
Quickly replying to just one part of that (I’ll get back to the rest later).

If ObsoleteAttribute doesn’t already add the deprecated directive it should. I have no issue having documented rules of inbuilt attributes mapping to graphql feature where is logically  makes sense. Like DescriptionAttribute is used.

bzbetty:
realised after a sleep i just need a class like DirectiveProcessor that detects attributes and registers the relevant directive, should work fine with build in attributes.

bzbetty:
added in a way for directives to register themselves based on built in attributes.

not happy with sheer number of classes that have been added, might ponder if I can simplify this somehow.


next step was to try call the directives from the introspection class purely to set the IsOneOf or IsDeprecated properties (would mean we wouldn't need the props on the SchemaTypes although not sure that's worth it) 

Main goal was to basically see if I could remove the concept of IsOneOf into the directive so rest of the codebase didn't need to know about it, maybe that's trying too much.

bzbetty:
Tried something a bit new.

* renamed to ExtensionAttribute, added second method to ApplyExtensions to ISchemaType and IField (was Field - this may have implications)
* Added second interface: IExtensionAttribute (which is probably named wrong), this is to wrap existing dotnet attributes to provide a similar API to FieldExtensionAttribute
* 
```
public interface IExtensionAttribute<T, A>
{
  public abstract void ApplyExtension(T field, A attribute);
}

public class ObsoleteAttributeRegistration : IExtensionAttribute<IField, ObsoleteAttribute>
    {
        public void ApplyExtension(IField field, ObsoleteAttribute attribute)
        {
            field.Deprecate(attribute.Message);
        }
    }
```

* Moved Deprecate method off IField to an extension method instead
* Removed IsDeprecated, DeprecatedReason and IsOneOf properties from IField and ISchemaType
* Added ISchemaDirective which is now responsible for
** Updating Introspection model fields (IsDeprecated, Reaosn and IsOneOf) - reason being that I suspect these fields will go away one graphql introspection supports schema directives, and i wanted all code related to these directives in one spot
** Rendering the directive itself in SDL - means we don't need to pass the arguments around
** Validate the directive is on the right type

* Potential Fix - not sure inline arguments were working for mutations at all (although i suspect this change is the reason i have a bunch of broken unit tests atm)
eg 
```
 mutation Test {
  createOneOfInputType(input: { one: 1 })
}
```

* Added concept of events.
** ISchemaType now has OnFieldAdded and OnValidate
** Could have added these as extra methods on the directive but thought it might be more flexible to use events instead (eg other things would be able to hook into these events too)
** These were both added to provide the OneOf directive validation meaning the whole implementation of OneOf is now basically in one file.

```
   public static void OneOf(this ISchemaType type)
        {
            type.AddDirective(new OneOfDirective());

            type.OnAddField += (field) => { 
                if (field.ReturnType.TypeNotNullable)
                {
                    throw new EntityQuerySchemaException($"{type.TypeDotnet.Name} is a OneOf type but all its fields are not nullable. OneOf input types require all the field to be nullable.");
                }
            };

            type.OnValidate += (value) => {
                if (value != null)
                {
                    var singleField = value.GetType().GetProperties().Count(x => x.GetValue(value) != null);
                    
                    if (singleField != 1) // we got multiple set
                       throw new EntityGraphQLValidationException($"Exactly one field must be specified for argument of type {type.Name}.");
                }
            };

        }
```

* New ApplyAttributes method on SchemaType and Field
** Concept is to group a lot of the functionality added to types/fields from attributes in one place and allow for extensibility (directives and extensions).  There's others that could potentially move there too (eg DescriptionAttribute, which could potentially be a IExtensionAttribute)

lukemurray:
Cool I like the new direction. We can still have extension methods to support the fluent style `type.AddField(...).IsDeprecated()` etc I assume?

Are the events required for the directive changes?

bzbetty:
> Cool I like the new direction. We can still have extension methods to support the fluent style type.AddField(...).IsDeprecated() etc I assume?

Yes the OneOf and Deprecated functions are extension methods now located with the attributes themselves.

> Are the events required for the directive changes?
Yes the two events added are for supporting the OneOf attribute.


I still need a way to turn these on/off instead of doing assembly scanning, for both performance and customisabilty sake.  eg I'm not sure everyone would want the @authorize directive (which honestly I'm not sure if it even belongs in this or should be something I have in my project, or even just a sample in the documentation on how to build one).

lukemurray:
`OneOf` I'm fine being in as it seems very close to being spec. `authorize` I'm not sure about? If not I'd rather it live in your codebase for now. But it could be a great example in the docs on how to build a custom schema-directive!

I'm not sure I follow this bit
>I still need a way to turn these on/off instead of doing assembly scanning, for both performance and customisabilty

Schema set up is typically done once. Not every time so performance is always important but less key than something that happens on execution. What does the events let you do? (don't get me wrong I think they could be useful, I'm wondering are they a different discussion/PR)

bzbetty:
> I'm not sure I follow this bit
>>I still need a way to turn these on/off instead of doing assembly scanning, for both performance and customisabilty

> Schema set up is typically done once. Not every time so performance is always important but less key than something that happens on execution.

Currently in the ApplyAttributes method I'm scanning all assemblies for compatible types, it would make sense to either do this once and cache the value somewhere (performance) and/or allow the user to add them manually.  That way we have the option of having some in the codebase which aren't enabled by default, which would also work better with unit tests as we could limit what is being applied.

I'll move the authorize directive to unit tests and reference them as an example.

> What does the events let you do? (don't get me wrong I think they could be useful, I'm wondering are they a different discussion/PR)

There's only two events added in this PR, both are needed by the OneOf directive as per below (OnAddField and OnValidate).
```

 public static void OneOf(this ISchemaType type)
        {
            type.AddDirective(new OneOfDirective());

            type.OnAddField += (field) => { 
                if (field.ReturnType.TypeNotNullable)
                {
                    throw new EntityQuerySchemaException($"{type.TypeDotnet.Name} is a OneOf type but all its fields are not nullable. OneOf input types require all the field to be nullable.");
                }
            };

            type.OnValidate += (value) => {
                if (value != null)
                {
                    var singleField = value.GetType().GetProperties().Count(x => x.GetValue(value) != null);
                    
                    if (singleField != 1) // we got multiple set
                       throw new EntityGraphQLValidationException($"Exactly one field must be specified for argument of type {type.Name}.");
                }
            };

        }
```

They were needed to implement all the oneof functionality that was previously implemented (but using the new approach).

Fleshing out these events more would certainly be in another PR.

bzbetty:
* merged from master
* flattened history as it was getting a bit messy.
* move authorize directive to asp.net tests project

still need to 
* add documentation 
* update apply attributes to not assembly scan, will probably just do it once in SchemaProvider and cache the list of types

lukemurray:
Conflicts again :)

Hopefully get some time to give this a good look at soon to bring into master branch and prepare for a 4.0 release.

bzbetty:
fixed conflicts

bzbetty:
Any thoughts on doing a prerelease alpha on nuget so we can easily test it against projects?

lukemurray:
Yep. Was going to do a 4.0-beta once we merge this in. 

Think I should do one earlier?

bzbetty:
That sounds fine, just noticed the lack of 3.0 beta was all.

docs/docs/directives/schema-directives.md
```diff
@@ -0,0 +1,133 @@
...
+# Schema Directives
+
+Schema Directives are used to decorate the schema to provide more information to clients, EntityGraphQL can also use this information to update the introspection model or provide extra functionality such as validation.
+
+## Built-In Directives
+
+The GraphQL spec defines directives that are supported out of the box in EntityGraphQL.
+
+- `@deprecated(reason: String)` - Tells the client that this type, field or enum value should no longer be used along with the reason why or a suggested alternative
+- `@oneOf` - Skip this field if the argument is true.
```
lukemurray, line 14: Copy and paste error with the description of this one and the next.

Can we also link to the `oneOf` proposed spec (https://github.com/graphql/graphql-spec/pull/825)

docs/docs/directives/schema-directives.md
```diff
@@ -0,0 +1,133 @@
...
+        field.Deprecate(attribute.Message);
+    }
+}
+```
+
+The `ISchemaDirective` method contains:
+
+* The `On` property for marking where it is valid to use this directive
+* The `ToGraphQLSchemaString()` method where you return the formatted directive including arguments (eg `return $"@deprecated(reason: \"{Reason}\")";`)
+* A series of methods the SchemaInstrospection class calls allowing you to modify the introspection result (primarily used because instrospection for @deprecated and @oneOf is done using specific fields and not yet in an extensible way)
```
lukemurray, line 82: Can you explain this comment more please?

src/EntityGraphQL/Schema/BaseField.cs
```diff
@@ -64,6 +65,39 @@ protected BaseField(ISchemaProvider schema, ISchemaType fromType, string name, s
             AddValidator<DataAnnotationsValidator>();
         }
 
+        public void ApplyAttributes(IEnumerable<Attribute> attributes)
```
lukemurray, line 68: When is this called? On schema set up or at compile time?

src/EntityGraphQL/Schema/BaseField.cs
```diff
@@ -64,6 +65,39 @@ protected BaseField(ISchemaProvider schema, ISchemaType fromType, string name, s
...
+        {
+            if (attributes.Any())
+            {
+                foreach (var attribute in attributes)
+                {
+                    if (attribute is ArgumentValidatorAttribute validator)
```
lukemurray, line 74: Any reason this needs to be a special case and not just another `extension.ApplyExtension(this)`?

src/EntityGraphQL/Schema/BaseSchemaTypeWithFields.cs
```diff
@@ -40,6 +45,35 @@ protected BaseSchemaTypeWithFields(ISchemaProvider schema, string name, string?
             RequiredAuthorization = requiredAuthorization;
         }
 
+        public void ApplyAttributes(IEnumerable<Attribute> attributes)
```
lukemurray, line 48: This looks the same as the one above. Can we refactor out to a single method in a helper/extension or similar?

src/EntityGraphQL/Schema/BaseSchemaTypeWithFields.cs
```diff
@@ -121,6 +153,30 @@ public void RemoveField(string name)
...
 
+        public ISchemaType AddDirective(ISchemaDirective directive)
+        {
+            if (
+                (GqlType == GqlTypeEnum.Scalar && !directive.On.Contains(TypeSystemDirectiveLocation.SCALAR)) ||
+                (GqlType == GqlTypeEnum.Object && !directive.On.Contains(TypeSystemDirectiveLocation.OBJECT)) ||
+                (GqlType == GqlTypeEnum.Interface && !directive.On.Contains(TypeSystemDirectiveLocation.INTERFACE)) ||
+                (GqlType == GqlTypeEnum.Enum && !directive.On.Contains(TypeSystemDirectiveLocation.ENUM)) ||
+                (GqlType == GqlTypeEnum.Input && !directive.On.Contains(TypeSystemDirectiveLocation.INPUT_OBJECT))
+            //todo (GqlType == GqlTypeEnum.Union && !directive.On.Contains(TypeSystemDirectiveLocation.UNION) ||
```
lukemurray, line 164: Should this be uncommented?

bzbetty: yep

src/EntityGraphQL/Schema/BaseSchemaTypeWithFields.cs
bzbetty: couple of minor differences

* field applies validators
* One runs applyextension using IField are the other ISchemaType arg

could probably be merged, wanted to figure out a plan for the assembly scanning first.

docs/docs/directives/schema-directives.md
lukemurray: I mainly meant the "not yet in an extensible way" part. Which bit is not yet extensible? 

It says `@deprecated` is the original comment but you mention `@deprecated` has a "nice reusable syntax" above?

Just trying to understand what this means? Each directive seemed to implement their own functionality, am I missing something?

src/EntityGraphQL/Schema/BaseField.cs
lukemurray: isn't that fine? The directive just doesn't override the type method?

docs/docs/directives/schema-directives.md
bzbetty: "not yet in an extensible way" as in there's no official place to put information around new directives you make.  You have ot invent a new field and look it up yourself.

The @deprecated comment was more about the above sentence eg SDL @directive syntax supports any directive, but that information isn't officially exposed in introspection for custom directives.

src/EntityGraphQL/Schema/BaseSchemaTypeWithFields.cs
lukemurray: Ahh yes, missed those details. Worth looking at merging in some generic or other way, lots of shared code.

What is the assembly scanning happening here?

docs/docs/directives/schema-directives.md
bzbetty: should probably be changed to a sentence stating something like

'directives are automatically added to the SDL output. As there is no official way to view directives in the introspection spec as of yet, the recommend method is extending the built in introspection types with additional fields with the information you wish to expose.

eg the @oneOf directive is exposed as the field OneField on TypeElement"

lukemurray: Oh I get it now.

Umm yeh interesting. `OneField` is nearly in spec. hopefully they bring in directive introspection into spec too.

Your above sentence is more to the point I think.

src/EntityGraphQL/Schema/BaseSchemaTypeWithFields.cs
lukemurray: OK I understand the scanning happening here now.

I know is is not as automatic (can have some defaults) but I feel it might be more easier to follow if we just register these? Like field extensions. 

I.e. 

```cs
schema.AddAttributeHandler<AuthorizeAttribute>();
```

That is done before any `SchemaBuilder` reflection, then it might be easier to follow what is happening.

Can register a few by default - Obsolete, etc.

Thoughts?

bzbetty: yeah actually i quite like that idea.

PR #196, opened 2022-08-06
Initial subscriptions implementation/support

### LINKED ISSUE #9: Implement GraphQL subscriptions

lukemurray:
# Implementation of GraphQL Subscriptions.

Fixes #9 

- Uses the dotnet `IObservable<>` / `IObserver<>` interfaces to implement the pub/sub functionality generically
- Supports the `graphql-ws` [protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) - This is a popular protocol for graphql subscriptions and supported by client libraries like Apollo

**Check out the example under `src/examples/subscriptions/`**

## Key classes in `EntityGraphQL.AspNet`

Note that the GraphQL spec doesn't state how to deliver the updates/stream. I have implemented a popular web socket one. Hence a chunk of the implementation lives in the ASP.NET package.

`GraphQLWebSocketServer.cs` - implements the `graphql-ws` [protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) 

`WebSocketSubscription.cs` - This class wraps the base subscription classes with the web socket implementation to stream events.

## What is happening?

1. Client sends a subscription operation. E.g.

```gql
subscription ChatRoom {
      onMessage { id user text timestamp }
}
```

2. EntityGraphQL compiles that and executes the registered `OnMessage` subscription. This is registered in the schema like mutations are and needs to return a `IObservable<>`

```c#
// add to schema
schema.Subscription().AddFrom<ChatSubscriptions>();

// implementation
public class ChatSubscriptions
{
    [GraphQLSubscription("Example of a subscription")]
    public IObservable<Message> OnMessage(ChatService chat)
    {
        return chat.Subscribe();
    }
}
```

3. `GraphQLSubscriptionStatement` executes the above and will return a `SubscriptionResult` which has the return value of the above method (`IObservable<>`) and the `GraphQLSubscriptionStatement` and `GraphQLSubscriptionField` data need to build the results on each new event/data

4. Client is waiting for `error` or `next` events

5. This is where `GraphQLWebSocketServer` comes in. It  has the result of executing the subscription statement/operation and sets things up. It creates `WebSocketSubscription` which ties the data (by subscribing to the `IObservable<>` to the socket.

6. When data comes through (via. the `IObserver<>` which is the `WebSocketSubscription`) we send the data to the client via the connected web socket

## Things to help in your services

`Broadcaster<TType>`  - Simple class to broadcast messages to all subscribers. You don't need to use it but it helps. Usage:

```c#
public class ChatService
{
    private readonly Broadcaster<Message> broadcaster = new(
    
    public void PostMessage(string message, string user)
    {
        // ... do your logic
        // broadcast the new message/event to all subscriber
        broadcaster.OnNext(msg);
    }
    // return the broadcaster as an IObservable<>
    public IObservable<Message> Subscribe()
    {
        return broadcaster;
    }
}
```

Keen for feedback, testing etc.

## TODO

- [ ] Lots of tests
- [x] Documentation
- [x] support query & mutation statements over web sockets as per protocol spec


lukemurray:
@bzbetty @breyed Not sure if you're interested in subscriptions or maybe have experience with web sockets and/or `IObservable<>`/`IObserver<>` but I'm keen to get some more eyes on this implementation and you've both been active here 🙂

breyed:
Why IObservable rather than  IAsyncEnumerator?

lukemurray:
I’ve not worked with that interface but a quick look it doesn’t provider pub/sub it is just an interface to provide the async enumeration. Which is one part but in this case you may have many clients subscribing to the same source (onMessage for example) so we need a way to manage those subscriptions and publish data to them.

lukemurray:
Looking at `IAsyncEnumerator` it is a pull style interface vs. push.

Without implementing something I'm not sure how it would work. The main idea here is that somewhere (possibly many places) you publish an event for a subscription. I.e. in the example in this PR the `ChatService` publishes events on new messages which is triggered by some client using the mutation to post messages. If no one is subscribed that is fine. If 20 are subscribed, they all get notified. 

With `IAsyncEnumerator`, `ChatService` would have to `yield` results? What happens if no one is currently trying to "pull" data from the enumeration (no subscribers)?

src/EntityGraphQL.AspNet/WebSockets/WebSocketSubscription.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.Collections.Generic;
+using EntityGraphQL.Compiler;
+
+namespace EntityGraphQL.AspNet.WebScokets
```
bzbetty, line 5: namespace spelling

bzbetty:
Overall looks like a fairly elegant implementation.

Haven't had a need for subscriptions, but I must admit I'd love to play with them.

I'd have a bit of work to figure out how to get them into my current project before i could do a proper test (guessing I wouldn't be able to use the current websocket implementation since I run in Azure functions and would need to introduce Web PubSub or something).

Like the new Mutation().AddFrom() method, should start to clean things up.

lukemurray:
Thanks @bzbetty.

Was there much different to get things working (query and mutations not subscriptions) on Azure functions? Been keen to document a getting started guide for functions if it is not so straight forward

bzbetty:
> Thanks @bzbetty.
> 
> Was there much different to get things working (query and mutations not subscriptions) on Azure functions? Been keen to document a getting started guide for functions if it is not so straight forward

Not a lot of effort, i mainly just used the 'Custom Controller / Manual Execution' section of the documents with a couple of small tweaks (manually deserializing the query so i could have direct accecss to the httprequestmessage still and a bit of code to deal with authentication - which are both fairly standard issue for people used to Azure Functions)

breyed:
My understanding is that IObservable is push, IEnumerable is pull, and IAsyncEnumerable is a combination. The main differences between IObservable and IAsyncEnumerable are that IAsyncEnumerable has integrated async/await/yield support, uses cooperative flow control, and is created per subscription. IObservable, by contrast, uses event-style coding, doesn't have flow control, and is created per source.

For IObservable, the source calls OnNext whenever an event occurs. The subscriber is responsible for keeping up with events or some buffer in between is required. For IAsyncEnumerable, the publisher is only allowed to send the next event in response to a call to IAsyncEnumerator<T>.MoveNextAsync(). It's still push in that if there is no new event yet when MoveNextAsync, the publisher will wait until an event occurs and then push the event via Task.SetResult. If an event occurs while a MoveNextAsync task is not pending, the publisher needs to buffer the event.

In the context of EntityGraphQL, the canonical example of a subscription would be extending a database query to return existing query results initially followed by future query results as they happen. Ideally, EntityGraphQL will support providing query updates automatically based on the GraphQL schema, plus support custom observable results that aren't extensions of DB queries.

IAsyncEnumerator makes a convenient common interface. The initial IQueryable result can be automatically converted via [AsAsyncEnumerable](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.entityframeworkqueryableextensions.asasyncenumerable?view=efcore-6.0). Custom initial results can be provided with yield statements. For future real-time updates, the DB query version can use a database eventing API (e.g., in Postgres, creating a trigger to call `notify` and then using [listen](https://www.npgsql.org/doc/wait.html) to cause a re-query for the updates). A custom subscription result can manually yield additional results as they occur.

Another advantage of IAsyncEnumerable is it allows the database to act as the buffer. Suppose that between while update A was in flight to the subscriber, 2 new rows were inserted. After the subscriber processes the results from update A, it will call MoveNextAsync, which will get EntityGraphQL going on update B. EntityGraphQL can now re-query the graph, using `where` clauses to limit results to those updated since the query for update A. If there are multiple subscribers, each will have its own state keeping track of where it last left off.

To facilitate automatic generation of incremental queries, the schema can identify a monotonic "updated at" column for each table.

lukemurray:
Thanks for that @breyed. I need to do more reading on `IAsyncEnumerable`/`IAsyncEnumerator`. And maybe look at trying to implement it when I get time again.

My quick read didn't lead me to believe it fit the pub-sub model. More use when something wants to stream data from a remote place. Which works for streaming a query result from a DB etc. But there is also the case where the data may not come from the DB. It may be an event triggered by a mutation. i.e. the data is not in one place to stream from it is an event that is triggered.

I'm probably just thinking about it wrong. I'mm dig into `IAsyncEnumerable`/`IAsyncEnumerator` when I get a chance.

breyed:
There are at least 3 use cases at play here:
* A large, slow query that is desirable to consume piecemeal (this use case is probably least common)
* A live query (existing results followed by updated results as updates are made)
* Event publication not related to a database query

Each use case seems to require a separate GraphQL feature. Either interface, IAsyncEnumerable or IObservable, can make any of the 3 use cases work. It's a matter of what's easier to use and most efficient.

Some interesting articles I found: [graphql-live-query](https://github.com/n1ru4l/graphql-live-query), which mentions support for both interfaces (their JavaScript equivalents actually, that is, AsyncIterable and Observable); the "Live Queries over Subscriptions" section of [this post](https://www.the-guild.dev/blog/subscriptions-and-live-queries-real-time-with-graphql), which shows an example of yielding initial results followed by real-time results; and the [PostGraphile](https://www.graphile.org/postgraphile/live-queries/) and [Hasura](https://github.com/hasura/graphql-engine/blob/master/architecture/live-queries.md) live query docs, which discuss approaches to tracking database updates.

lukemurray:
Agree with "Each use case seems to require a separate GraphQL feature". This PR is purely looking at the subscription spec.

Live queries I want as well and there seems to be a few community driven implementations as it is not in the spec.

The current spec covers the 2 & 3 is a generic sense. A client subscriptions to some defined event stream and gets updates. You could implement a "live style" query by first doing a normal `query` then `subscription` to an event stream of updates. And point 3 is the classic subscription to some event topic.

So first step was to cover the spec. Again which actually says nothing about the transport (web sockets or not) but graphql-ws is well supported by tools. Then next steps look at things like live queries etc.

lukemurray:
@breyed Not sure if you have ideas what an `IAsynEnumerable` implementation might look but I struggled to get something working that doesn't feel like I'm jumping through hoops. I'm likely missing something as the documentation for them in dotnet is near nothing and the examples online barely go past a simple `async` for loop. And looking at the EF Core source code (seems the biggest place they are used) has helped a bit but still not getting anywhere.

I'm not looking at live queries (yet) or streaming data to the client for this. I think `IAsyncEnumerable` might fit better there.

For this I'm looking at the spec GraphQL subscription where
- the client registers a subscription using the `subscription` operation
- The server sends updates/events for that subscription event

And using the graphql-ws protocol as it is well supported by tools. This means
- Each client creates a web socket connection to the server
- each client may register multiple subscriptions over that single web socket connection
- The web socket needs to be available to receive messages like `ping` and other `subscribe` messages from the client

I also want to support subscriptions fairly generically so events could come from anywhere
- data changes (you'd have to for them outside of EntityGraphQL)
- some trigger in a mutation or other API (i.e. on data change or post message in the example in this PR)
- sign in events or other events from other services
- basically anywhere you may want to trigger an event

This means we need an easy way to trigger "this event happened with this data" from many places (this is `IObservable` and a simple implementation `Broastcaster`) and if there are any subscriptions to that event we send data to the client (matching their field selection). Pub/sub (`IObservable`) fits very well here.

With `IAsyncEnumerable` each time a `subscribe` event comes over the web socket it executes it (e.g. `ChatSubscriptions.OnMessage` which returns an `IAsyncEnumerable`.

We then need to hold something that is doing a `async` for loop over that `IAsyncEnumerable` which is basically waiting for data. However there may be 20 clients subscribed to the same event meaning 20 for loops over the same `IAsyncEnumerable`. Which I don't believe will work.

So we need to introduce something that feeds the event trigger into many `IAsyncEnumerable` (one for each subscription). Which feels like I'm building pub/sub under `IAsyncEnumerable`.

On top of that I couldn't figure out a nice way to have the `IAsyncEnumerator` wait until there is data. `ValueTask<bool> MoveNextAsync()` needs to wait and resolve `true` when there is data (or `false` to end the loop). I can't see a nice way to have it wait and have it resolve `true` once we have data and have set `Current` to be read and then make it wait again.

You can see what I tried in the branch `feature/subscriptions-async` and see a diff of it between this PR here https://github.com/EntityGraphQL/EntityGraphQL/compare/feature/subscriptions...feature/subscriptions-async?expand=1

Keen for thoughts, feedback, etc on that. It "works" in the simple chat example I have. But I think requires a lot of testing etc and I feel that must be a better way to do it. The `IObservable` way feels like a cleaner implementation for this use case.

breyed:
It seems that each interface has its place. For cases where a single source object is providing updates to multiple subscribers, IObservable fits well. For cases where a separate source object is created for each subscriber (analogous to per-request ASP.NET controller objects and EntityGraphQL mutation objects), IAsyncEnumerable fits well.

In the IAsyncEnumerable case, there is no need for EntityGraphQL to wrap the interface. It can use it directly:
```c#
await foreach (var update in feed) {
   // Send update over WebSocket
}
```

The GraphQL context likewise can use C# support for IAsyncEnumerable provide the data, e.g. for a live query that gets future results via another IAsyncEnumerable:
```c#
var futureResults = SubscribeToFutureResults();
foreach (var result in QueryExistingResults())
    yield return result;
await foreach (var result in futureResults)
    yield return result;
```

lukemurray:
Yes the implementation I posted does the `await foreach` [here](https://github.com/EntityGraphQL/EntityGraphQL/compare/feature/subscriptions...feature/subscriptions-async?expand=1#diff-93fa7d9a79b4bf0c4149be55e0e53ae03217fa4eb99729730ae72f722fa29e05R50)

But the `feed` needs some implementation of `IAsyncEnumerable`. That implementation provides a simple [`SubscriptionEnumerable`](https://github.com/EntityGraphQL/EntityGraphQL/compare/feature/subscriptions...feature/subscriptions-async?expand=1#diff-56401dc106e1c5ae9550898990fc6f8cbde147826f5b0b0486665f90ef0fd933R26) to do that but we need to create 1 of those for every subscription. Which works just feels a lot of extra effort compared to `IObservable<T>`.

I agree they have their places and you put it well, and I think it'll fit better with live queries.

breyed:
If I understand correctly, the feed you're referring to that needs to implement IAsyncEnumerable is the same as the GraphQL context I provided sample code for in my last comment. In a method that returns IAsyncEnumerable, the compiler does all the heavy lifting for the implementation automatically. The code that a user of EntityGraphQL writes would be pretty simple, like:

```c#
class MyGraphQLContext {
    public async IAsyncEnumerable<Result> Results => {
        var futureResults = SubscribeToFutureResults();
        foreach (var result in QueryExistingResults())
            yield return result;
        await foreach (var result in futureResults)
            yield return result;
    }
}
```

A downside of this approach, however, is that it would fetch entire entities from the database, even if the GraphQL client only requested a few fields. In line with EntityGraphQL's core differentiating feature, an improvement is to limit which columns are queried from the database. In other words, how do we combine the best of IQueryable and IAsyncEnumerable? For the live query use case, a simple solution is to break the two components into separately returned components. Return a tuple consisting of immediate results and future results. The `Results` property in the above example would become:

```c#
    public async (IQueryable<Result>, IAsyncEnumerable<Result>) Results =>
        (QueryExistingResults(), SubscribeToFutureResults());
```

Where the types of the source methods are the same as before:
```c#
    IQueryable<Result> QueryExistingResults();
    IAsyncEnumerable<Result> SubscribeToFutureResults();
```

EntityGraphQL can use its existing query optimization for the existing results. In many use cases, the queries for existing items will account for the vast majority of the data, making the optimization very helpful. The subscription for future results would always provide the entire entity to EntityGraphQL. That's not ideal, but I don't know of a good alternative. Fortunately, EntityGraphQL will still optimize how much data gets sent back to the client.

lukemurray:
What I meant was the `SubscribeToFutureResults` methods needs to return something that implements `IAsyncEnumerable`. The user could build something or use the `SubscriptionEnumerable` in that PR. Note that EF implements that interface in `QueryingEnumerable` https://github.com/dotnet/efcore/blob/0577a3acab690c9d884889d00e56f3743edb52c4/src/EFCore.InMemory/Query/Internal/InMemoryShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs

I didn't use that in the example PR as these PRs are not focused on live queries but a generic sub/sub subscription implementation. We might be able to use some of those EF features when we look at live queries - but still need a way to support service fields too.

Right now for the spec GraphQL subscription feature that is not tied to EF or anything I think this PR provides that.

breyed:
I can't find where `SubscriptionEnumerable` is defined, but in any case, I expect often the `IAsyncEnumerable` will be come from `yield return` statements, typically in a for loop. Sometimes, a `ToAsyncEnumerable` extension method in `System.Linq.Async` will be helpful to convert from `IObservable` or a non-async collection, especially if combining from various underlying sources.

It would be helpful for API design to have a few sample concrete use cases. From there, you could see what makes for the cleanest application code, and design the EntityGraphQL API to suit.

lukemurray:
Sorry there is a branch `feature/subscriptions-async`

You can see the diff here https://github.com/EntityGraphQL/EntityGraphQL/compare/feature/subscriptions...feature/subscriptions-async?expand=1#diff-56401dc106e1c5ae9550898990fc6f8cbde147826f5b0b0486665f90ef0fd933R26 - ctrl-f for SubscriptionEnumerable.

I didn't create a PR. I.e. it is a different branch from the one in this PR.

My understanding of things like `ToAsyncEnumerable` is it turns a list into async. In subscriptions we do not have a known/resolved list as there may be events that come in hours or days later.

`SubscriptionEnumerable` is basically waiting for events and then feeding the `await foreach`. You can see in that branch `WatchForEvents` is doing the `await foreach` and there is code to handle getting an event and pushing it to many subscriptions.

The GraphQL spec leaves subscriptions fairly generic and very pub/sub focused. Users subscribe to an event (subscription field) and provides the selection graph - the data they want sent to them. The events could be from anywhere. A user signs in. A user signs up. New data somewhere, deletion of data etc.

Use cases we have are
- user signs in
- user is now viewing a certain page (other users on that page get notified)
- user signs out
- changes have been made to a object graph (live queries can solve this in a different way)
- New data in a shared cache (not in our DbContext DB

breyed:
`SubscriptionEnumerable` is similar to [ToAsyncEnumerable](https://github.com/dotnet/reactive/blob/main/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.Observable.cs), which converts an `IObservable` to an `IAsyncEnumerable`, with the biggest difference being the `ConcurrentQueue` in `ToAsyncEnumerable`. Since `SubscriptionEnumerable` doesn't do any buffering, if a second call is made to `PushNext` before `MoveNextAsync` is called, `tsc.SetResult` will throw an exception.

Buffering is the key factor in analyzing the use cases. For example, for the user sign-in use case, suppose a client subscribes, but then the server it had connected to goes down. The client detects this and establishes a new subscription, this time getting another server in the cluster. In the meantime several users had signed in (using working servers). Is it important that the server inform the client about the sign-ins it had missed? The answer will vary by application.

One type of application is real-time best-effort. A server provides events that it is aware of at a rate the client can accept. Events during downtime or too fast for the client are dropped.

Another type of application is guaranteed transactional history. In this case, a client subscribes at some time T0 and is guaranteed to eventually receive all events between T0 and some T1 when it unsubscribes. This kind of application needs to persistently store events so that they can be re-queried in case a client's server (at the time) was down when the event occurred.

For either type of application, a server can benefit from in-memory buffering. In the best-effort case, it reduces event loss due to transient network slowness. In the transactional history case, it serves as a cache for the same scenario so that the event database doesn't need to be re-queried.

The scenarios drive the GraphQL API. For real-time best-effort, IObservable is simplest because it's push-and-forget. It can have some internal buffer. If it's exceeded, excess events are dropped, but that's OK. For transactional history, you essentially always need a live query. There, IAsyncEnumerable is better. It ensures events aren't pushed faster than the client can handle, which is important because dropped events are not OK. Instead, if an event arrives in real time and the client is ready, it pushes the event (by completing the task of a MoveNext). If not, when the client is ready, the server will iterate over missed events from the persisted store (or in-memory cache if available).

lukemurray:
`SubscriptionEnumerable` is more like the `ObservableAsyncEnumerable` that `ToAsyncEnumerable` returns. Also that is in another library that EntityGraphQL doesn't yet have a dependency on, although maybe it should. I'll look more into that library.

Interestingly the first cut at `SubscriptionEnumerable` also used a queue to manage data.

The buffering stuff is interesting and likely a large impact. If a client subscribes and never unsubscribes but looses connection you do not want to buffer for ever. It could also get crazy with a single event stream having subscribers all at different positions because of their connection issues. Also a pure in-memory buffer doesn't work well with load balancers. So we'll need to provide ways to implement other buffer stores.

PR #201, opened 2022-08-12
Mutation autoAddInputTypes ignored without FromObject

breyed:
The `autoAddInputTypes` parameter for `AddMutationsFrom` doesn't seem to have any effect if the schemaProvider wasn't created with `SchemaBuilder.FromObject`.

The unit test result shows `Person.id` missing, ~but I suspect there are more problems beyond that. I came across this by noticing in my own code that the arguments to my mutation were missing, as if autoAddInputTypes had not been set to true.~

~A related problem is that EntityGraphQL silently removed the arguments from the mutation, rather than reported an error.~

### DIFF
--- a/src/tests/EntityGraphQL.Tests/MutationTests/MutationMethodParameterTests.cs
+++ b/src/tests/EntityGraphQL.Tests/MutationTests/MutationMethodParameterTests.cs
@@ -118,7 +118,44 @@ namespace EntityGraphQL.Tests
         }
 
         [Fact]
-        public void TestSeparateArguments_AutoAddInputTypes()
+        public void TestSingleArgument_AutoAddInputTypes_NestedType() {
+            var schemaProvider = SchemaBuilder.FromObject<TestDataContext>();
+            schemaProvider.AddMutationsFrom<PeopleMutations>(true, true);
+            // Add a argument field with a required parameter that is defined in a nested class
+            var gql = new QueryRequest {
+                Query = @"mutation AddPersonSingleArgumentNestedType($nameInput: InputObject) {
+                  addPersonSingleArgumentNestedType(nameInput: $nameInput) { id name }
+                }",
+                Variables = new QueryVariables {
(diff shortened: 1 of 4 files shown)


breyed:
It looks like the issue with removed arguments (in my testing on a production project, not in the unit tests) was caused by #203.

breyed (commented):
The tests in commit 434d46ed4524f41d3228417eea0e135049fe05b3 are unrelated to auto-adding input types. I should probably have put them in a separate PR. Hopefully, they don't cause any confusion being here.

bzbetty:
I think https://github.com/bzbetty/EntityGraphQL/pull/new/autoAddInputTypes-test fixes #201

also made a couple of type fixes to the test themselves once the bug was fixed.

bzbetty:
Actually that first fix might make a regression with dbcontext being an arg. Will have to make sure to check if there's a unit test on that one.

breyed:
> should mutation return types be auto registered

Since return types are not input types, `AddMutationsFrom` currently doesn't have a parameter to opt into automatic inclusion. However, it seems reasonable to adjust the meaning of `SchemaBuilderOptions.AutoCreateInterfaceTypes` to include mutation return types in addition to types in the object graph.

lukemurray:
Do you mean adjust the meaning of `SchemaBuilderOptions.AutoCreateNewComplexTypes` to include mutation return types?

Where that description is
>If true (default) and an object type is encountered during reflection of the object graph it will be added to the schema as a Type including it's fields. If that type is an interface it will be added as an interface

And `SchemaBuilderOptions.AutoCreateInterfaceTypes` is
>If true (default = false), any object type that is encountered during reflection of the object graph that has abstract or interface types (regardless of if they are referenced by other fields), those will be added to the schema as an Interface including it's fields

breyed:
Yes, except I hadn't thought about `AutoCreateInterfaceTypes`. I agree it makes sense to increase the scope of both properties to include mutation return types.

lukemurray:
See PR #208

PR #212, opened 2022-08-22
Can't implement service via static method.

breyed:
When adding a field implemented through a service, if the implementation is invoked through a static method, the call fails. This PR has a unit test `TestServiceFieldWithStaticMethod`, which results in this error:

> Field 'projects' - variable 'srv' of type 'EntityGraphQL.Tests.ServiceFieldTests+ConfigService' referenced from scope '', but it is not defined

On a production project with a similar configuration, the error I get is effectively:

> ''GetItems_entity, service_' is not a member of type 'Entity' (Parameter 'propertyOrFieldName')'

The reason I was using a static method in the first place is to subvert the use of `Expression`. There is complex business logic that combines data from the entity and service and returns a custom non-entity type. The expression tree is far too complex for LINQ. All I want to do is have the method executed and the entire object returned. If the GraphQL client only requested select properties, they would be cherry picked at the end, with the generation of the other properties wasted work (which is acceptable in this use case).

Note that this issue is a regression. The use of a static method works in EntityGraphQL 2.x.

### DIFF
--- a/src/tests/EntityGraphQL.Tests/QueryTests/ServiceFieldTests.cs
+++ b/src/tests/EntityGraphQL.Tests/QueryTests/ServiceFieldTests.cs
@@ -1608,6 +1608,54 @@ namespace EntityGraphQL.Tests
         }
 
         [Fact]
+        public void TestServiceFieldWithStaticMethod() {
+            var schema = SchemaBuilder.FromObject<TestDataContext>();
+            schema.AddType<ProjectConfig>("ProjectConfig").AddAllFields();
+
+            schema.Type<Project>().AddField("config", "Get project configs if they exists")
+                .ResolveWithService<ConfigService>((p, srv) => GetProjectConfig(p, srv)) // srv.Get(p.Id)) works here
+                .IsNullable(false);
+
+            var serviceCollection = new ServiceCollection();
+            var srv = new ConfigService();
+            serviceCollection.AddSingleton(srv);
+
+            var gql = new QueryRequest {
+                Query = @"{
+                    projects {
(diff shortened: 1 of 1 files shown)


breyed:
I refactored my production code to move make the static method an instance method of the GraphQL context, which I fed into `ResolveWithService` as a service. It also didn't work, so the issue isn't specific to whether the method is static.

What did make EntityGraphQL happy was using an instance method in isolation. My initial AddField lambda did an EF query and passed the IQueryable to the method, plus did a select on the method's output. Once I put all the logic into a method, so that AddField lambda was just `gql.GetResults(entity)`, when EntityGraphQL 3.0.5 worked.

lukemurray:
Interesting. I've never written it like this so I am not surprised it fails :)

Happy to (try) to make it work. Out of interest Does it give you anything over just using the service as below?

```cs
schema.Type<Project>().AddField("config", "Get project configs if they exists")
                 .ResolveWithService<ConfigService>((p, srv) => srv.DoSomething(project.Whatever))
                 .IsNullable(false);
```

Just a style thing? Or, just thought maybe you are using a service you can't change and want to wrap up functionality in that static method?

breyed:
The reason for the static in my case was just separation of concerns. The service wasn't a logical fit for adding the query code to. The Graph QL context was the logical spot for the query - ideally right in the call to AddField. But since the query wasn't just a simple expression, but required a statement block, it needed to be pulled out to a separate method.

lukemurray:
Yeh that makes sense. Tests added and fixed now - in master / unreleased currently. Thanks for the pick up.

Issue #205, opened 2022-08-13
Development and production error messages


breyed:
When a query in an added field fails, EntityGraphQL helpfully provides a GraphQL response that contains the error. However, the error message sometimes doesn't contain the message from the inner exception. For example, if Entity Framework Core is confirmed to throw on warnings in `OnConfiguring`, like this
```c#
optionsBuilder.ConfigureWarnings(b => b.Default(WarningBehavior.Throw));
```
EntityGraphQL doesn't show the actual error. For example, for a warning about client-side execution being required, the output is just:

```json
{
    "errors": [
        {
            "message": "Field 'customers' - Exception has been thrown by the target of an invocation."
        }
    ]
}
```

In development (per `IWebHostEnvironment`), the underlying cause should be shown. Conversely, for security in production, the error should show the same message that an ASP.NET Core API controller shows in production, with the exception message not provided.

lukemurray:
What error does ASP.NET Core API controller shows in production?

breyed:
I created a new project with:
```sh
dotnet new web
```

I changed the code to throw an exception:
```c#
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet("/", _ => throw new Exception("Sample error."));

app.Run();
```

I first left `ASPNETCORE_ENVIRONMENT` set to `Development` in `launchSettings.json`. In sent a GET request with the `Accept` header set to `application/json`. The HTTP response was a 500 Internal Server Error with this body:
```
System.Exception: Sample error.
   at Program.<>c.<<Main>$>b__0_0(HttpContext _) in /Users/brey/Code/Sandbox/aspnetcore/Program.cs:line 4
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

HEADERS
=======
Accept: application/json
Connection: keep-alive
Host: localhost:5247
User-Agent: PostmanRuntime/7.29.0
Accept-Encoding: gzip, deflate, br
Postman-Token: a5dab21c-1b37-4224-a4fa-e7ca5e2d1a9e
```

When I changed `ASPNETCORE_ENVIRONMENT` to `Production`, the HTTP response was the same 500 Internal Server Error, but this time the response had no body (`Content-Length` was 0).

[...]

PR #213, opened 2022-08-22
Unwrap multiple target invocation exceptions

breyed:
It is common for an exception from a field resolution method to be wrapped twice in `TargetInvocationException`. This PR unwraps as many `TargetInvocationException` layers as are present. It partially implements #205.

### DIFF
--- a/src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs
+++ b/src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs
@@ -100,18 +100,18 @@ namespace EntityGraphQL.Compiler
                     var errors = aex.InnerExceptions.SelectMany<Exception, string>(ex => ex is EntityGraphQLValidationException vex ? vex.ValidationErrors : new[] { $"Field '{fieldNode.Name}' - {ex.Message}" });
                     throw new EntityGraphQLValidationException(errors);
                 }
-                catch (TargetInvocationException ex)
-                {
-                    if (ex.InnerException is EntityGraphQLException vex)
-                        throw new EntityGraphQLException(fieldNode.Name, vex);
-                    throw new EntityGraphQLExecutionException(fieldNode.Name, ex.InnerException!);
-                }
                 catch (EntityGraphQLValidationException)
                 {
                     throw;
                 }
                 catch (Exception ex)
                 {
+                    while (ex is TargetInvocationException)
+                    {
+                        ex = ex.InnerException!;
+                        if (ex is EntityGraphQLException vex)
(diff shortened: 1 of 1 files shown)


src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs
```diff
@@ -100,18 +100,18 @@ public ExecutableGraphQLStatement(ISchemaProvider schema, string name, Expressio
...
-                        throw new EntityGraphQLException(fieldNode.Name, vex);
-                    throw new EntityGraphQLExecutionException(fieldNode.Name, ex.InnerException!);
-                }
                 catch (EntityGraphQLValidationException)
                 {
                     throw;
                 }
                 catch (Exception ex)
                 {
+                    while (ex is TargetInvocationException)
```
bzbetty, line 109: does GetBaseException not work in this case?
https://docs.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-6.0

breyed: It depends where in the exception chain is most valuable. Sometimes the base exception is too narrow. For example, suppose a service is fetching from an API and the network connection fails. The base exception may be a low-level SocketException. It may be wrapped in a HttpException, which itself may be wrapped by the service layer in a MyServiceException. All of this may then be wrapped in two layers of TargetInvocationException by EntityGraphQL. In this example, the MyServiceException would likely be the most valuable.

Since EntityGraphQL has the luxury of JSON for the output, perhaps it can provide an array or linked list of all the exception layers (except the non-useful ones like TargetInvocationException).

lukemurray: This fits the spec as well. A result (with errors) is actually a list of errors.

`QueryResult.AddError()` can be used. Currently there are a few places trying to catch exceptions and clean them up. Might be nice to bring that to one if we can.

`ExecutableGraphQLStatement` and `GraphQLMutationStatement` mainly catch exceptions to add `field.Name` to the detail.

In `SchemaProvider.ExecuteRequestAsync` it does some "unwrapping" should the logic in this PR be in that location to do the same for anything executed?

breyed: I found the location for the PR based based on the inspecting in the debugger the stack trace of a first-chance exception my code threw. I didn't look holistically to see where else unwrapping logic would be helpful. Factoring out unwrapping (or even better cataloging) exception layers into a common helper method makes sense to me.

PR #219, opened 2022-08-23
Required array parameter fails conversion

breyed:
When passing a required variable of an array type to a query, the conversion operator used to access the wrapped value inside RequiredField throws an exception.

In the unit test in this PR, the failed conversion occurs in the cast `(Guid[])args.ids`. However, the error also occurs in the more subtle case of an implicit conversion, such as if `args.ids` is passed to a method that takes a parameter of type `Guid[]`.

A downstream problem is that the error handling is throwing an NullRefernceException, causing the original exception related to `IConvertable` to not be output.


lukemurray:
Thanks for all the tests for cases we're missing!

breyed:
Happy to add the tests. Thanks for a library that is quite useful for complex enterprise applications (which are great at exposing edge cases). :-)

PR #223, opened 2022-08-25
bug: mutations with inline args don't support variables with different name to argument name

bzbetty:
```
  var gql = new QueryRequest
            {
                Query = @"mutation AddPersonSingleArgument($differentName: InputObject) {
                  addPersonSingleArgument(nameInput: $differentName) { id name }
                }",
                Variables = new QueryVariables {
                    { "differentName", new InputObject() { Name = "Frank" } },
                }
            };
```

was failing to find differentName as the argument for nameInput


lukemurray:
Does this fix #225 as well?

bzbetty:
> Does this fix #225 as well?

it might actually

PR #224, opened 2022-08-25
Hold the type a field  belongs to

lukemurray:
So we can better tell same name fields from different types apart for union types

@bzbetty thoughts on this instead of replacing the name to insert the type in one spot and removing it later?

Seems knowing the type the field came from would be useful elsewhere too.


bzbetty:
yah i really like that idea, didn't think of that.

PR #228, opened 2022-08-27
Cannot pass object with selected fields to resolution method

breyed:
To select fields from an object, EntityGraphQL creates a proxy object for it. This proxy object is incompatible with the original type. The incompatibility causes a runtime error when the proxy object is passed to a method used to resolve a subsequent field. This PR tests this scenario.

### DIFF
--- a/src/tests/EntityGraphQL.Tests/QueryTests/ServiceFieldTests.cs
+++ b/src/tests/EntityGraphQL.Tests/QueryTests/ServiceFieldTests.cs
@@ -704,7 +704,7 @@ namespace EntityGraphQL.Tests
             schema.AddType<ProjectConfig>("ProjectConfig").AddAllFields();
 
             schema.Type<Project>().AddField("config", "Get project config")
-                .ResolveWithService<ConfigService>((p, srv) => srv.Get(p.Id));
+                .ResolveWithService<ConfigService>((p, srv) => srv.Get(p));
             // because of the selections of the config.type (a service) when we do our second selection (first is EF compatible)
             // the type of t will not match, need to be replaced
             schema.Type<Project>().ReplaceField("tasks", p => p.Tasks.Where(t => t.IsActive), "Active tasks");
@@ -1841,6 +1841,8 @@ namespace EntityGraphQL.Tests
                     Type = "Something"
                 };
             }
+
+            public ProjectConfig Get(Project p) => Get(p.Id);
 
             public string Get(string str1, string str2)
             {


lukemurray:
I swear this was in the docs :)

Current recommendation is to pass in fully selected fields (like the test was doing with `p.Id`).

You can pass many

```cs
schema.Type<Project>().AddField("config", "Get project config")
    .ResolveWithService<ConfigService>((p, srv) => srv.Get(p.Id, p.OtherField, p.More));
```

This [page](https://entitygraphql.github.io/docs/entity-framework#how-entitygraphql-handles-resolvewithservice-with-ef) explains what is happening. Basically it first selects the `Id`, `OtherField` & `More` fields than can pass those into the service.

You should be able to do something like this too

```cs
schema.Type<Project>().AddField("config", "Get project config")
    .ResolveWithService<ConfigService>((p, srv) => srv.Get(new MyArgType { Id = p.Id, p.OtherField = p.OtherField, More = p.More }));
```

breyed:
Even though the "How EntityGraphQL handles ResolveWithService() with EF" you referenced is about EF, it looks like the creation of the proxy object is independent of EF, which is why the unit test, which doesn't use EF, fails when the full entity is passed to the helper method.

In the production case where I encountered this issue, I am using EF. I passed the entity to look up related records using the DB context. The database has a legacy structure with several multi-column primary keys, which means foreign key lookup using the navigation properties requires far less code. My equivalent to `Get` was like this:

```c#
from child in db.Children where child.Parent == parent
```

But that caused the proxy object error. Based on your feedback, I fixed the problem by passing and using each key column individually:

```c#
from child in db.Children where child.Parent.KeyColumn0 == parentKeyColumn0 && child.Parent.KeyColumn1 == parentKeyColumn1 && child.Parent.KeyColumn2 == parentKeyColumn2
```

The flow of the code is the same, as I suspect is the generated SQL. Thanks for pointing me in the right direction. If a clever solution comes to mind on day that would allow working with the full entities, it would be handy. For now, if there's no known practical way to do so, feel free to close this PR.

bzbetty:
It'd be possible to add implicit cast operators and maybe override equality function. Just not sure how that would translate to SQL. Just throwing ideas in 🙂

breyed:
It turns out that passing the key column values as separate parameters leads to a different problem:

The query is like this:
```graphql
query GetCustomers {
  customers {
      locations {
        services {
            id
        }
      }
  }
}
```

Where `services` is added to `Location` via `AddField`, with `Location.CustomerId` being one of the key fields. The result is a runtime error like this:
> "Field 'customers' - 'location_CustomerId' is not a member of type 'Location' (Parameter 'propertyOrFieldName')"

I couldn't repro this in unit test. (I tried using `TestDataContext.Projects.Children` to get an equivalent depth.) It may be the underlying problem is fixed. I was able to work around the issue by passing the entity to the service method (the `Get` equivalent) using `dynamic`. Thought I'd mention in case it's useful.

lukemurray:
Thanks. Yes I believe the issue you hit is fixed in the main branch here

lukemurray:
> It turns out that passing the key column values as separate parameters leads to a different problem

@breyed can you confirm if this second issue is resolved in 4.x?

breyed:
@lukemurray Yes, passing the key column values as separate parameters works in 4.0.0.

Issue #229, opened 2022-08-31
ResolveWithService vs Navigation Properties on other Fields


bzbetty:
I have two fields on a type that work if I query them separately, but if I ask for both in one query I get the error

"Field 'jobs' - 'LabourItems' is not a member of type 'Dynamic_items_2d5b8d61-83f3-44e0-b4dd-87b9edec3af9' (Parameter 'propertyOrFieldName')"

```
config.AddField("production", "").Resolve(j => j.LabourItems.Sum(li => li.DurationStd));                       

 config.AddField("start","").ResolveWithService<ESPContext>((job,db) => job.DueDate.AddWeekDays(-job.ProductionLead));
```

Currently using version 3.0.3 due to another issue in 3.0.5 that prevented me upgrading.

lukemurray:
Does this work with the master branch?

What is the issue in 3.0.5 that is preventing you from upgrading? I want to make sure we have test coverage for it.

Also, interesting in
```cs
config.AddField("start","").ResolveWithService<ESPContext>((job,db) => job.DueDate.AddWeekDays(-job.ProductionLead));
```
You are not using `db`, the service. EF might be able to translate `job.DueDate.AddWeekDays(-job.ProductionLead)` so you could use
```cs
config.AddField("start","").Resolve((job) => job.DueDate.AddWeekDays(-job.ProductionLead));
```

If EF cannot, then using `ResolveWithService` is correct.

bzbetty:
Can't upgrade to 305 due to #215 currently.

I'll try replicate this issue with a unit test.

bzbetty:
Using ResolveWithService since AddWeekDays isn't a method that Sql understands.  (if it was just add days I'd do it the other way)

PR #232, opened 2022-09-01
replicate & fix issue #229

bzbetty:


bzbetty:
not 100% on this one, but it seemed weird that the resolve function passed true for withServices.

lukemurray:
Good catch. It should be false. Otherwise it is incorrectly trying to extract fields.

PR #234, opened 2022-09-02
refactor registering attribute handlers

lukemurray:
@bzbetty any thoughts on these changes?


lukemurray:
`IExtensionAttributeHandler` lets you handle multiple types. E.g. the Authorise example you did. See here https://github.com/EntityGraphQL/EntityGraphQL/pull/234/files#diff-cf576b0b391a1cac85c5fdc79a689280d62e3efeebb12d12684b70f1ab04321eR7

The `AbstractExtensionAttributeHandler<T>` is just a helper that lets you work with types... may not be really needed?

Issue #235, opened 2022-09-06
Field arguments need unique names


bzbetty:
```
query {
  Jobs {
   Field1(arg: true) {}
   Field2(arg: false) {}
  }
}
```

Both fields get called with arg being false

Works fine on top level fields

bzbetty:
@lukemurray this doesn't quite cover the usecase I hit.

Arguments are still the same if you query the same subfield twice with an alias.

eg
```
 {
                people {
                    name
        	        project(id: 55) {
        		        name
        	        }
                    project2: project(id: 58) {
        		        name
        	        }
                }
            }
```

which is the one in my unit test I PR'd

lukemurray:
Ahh, thanks! I've added those tests to now.

PR #236, opened 2022-09-06
Replicate issue #235

bzbetty:


lukemurray:
Thanks, I'm actually in the middle of fixing this :)

PR #245, opened 2022-09-11
PoC: playing with neinlinq

bzbetty:
query {
  people { age }
}


bzbetty:
seems to play nicely with static methods.

instance I had to GraphQLIgnore the expression (otherwise it tried adding the type) and then it exploded with a bad argument rebind. didn't play around to see if that was a bug in EQL or NeinLinq.

A [GraphQLField] would work well.

PR #246, opened 2022-09-14
feat: documenting library compatiblity & BeforeExecuting Feature

bzbetty:


lukemurray:
@bzbetty this is still marked as draft. Is there a dot point list of what you'd like to see added/completed to finish this off?

bzbetty:
Was waiting on info about hooking into pre execution, but it sounds like you didn't have much luck there?

bzbetty:
so having a quick look at the preexecute hook,

would it not go in 

ExecutableGraphQLStatement.ExecuteExpressionAsync() 

between
`var lambdaExpression = Expression.Lambda(expression, parameters.ToArray());`

and

`Compile().DynamicInvoke`

Essentially a new property on ExecutionOptions ?

ExecutionOptions.OnBeforeExecute(LambdaExpression);

Thoughts?

lukemurray:
That is what I first though and quickly added it. Issue is that not all expressions have a `IEnumerable<>`/`IQueryable<>` result. Hence you can not cal any of the extension methods designed for that.

e.g. GraphQL

```gql
query Movie($id: ID!) {
   movie(id: $id) { name }
}
```

this creates a similar expression to this

```c#
ctx.Movies.Where(m => m.Id == args.id).FirstOrDefault()
```

You can not just wrap that in a method call expression that takes a `IEnumerable<>`/`IQueryable<>` as the argument. Depending on the schema configuration that expression might actually be more like

```c#
something == null ? null : new {
   field = something.Field
}
```

I haven't spent much more time thinking about it yet, but I think the few places we know we've finished a `IEnumerable<>`/`IQueryable<>` expression and are about to wrap in a `null` check or similar we can do the call back then.

bzbetty:
#253

bzbetty:
tested out ExpressionOptimizer a bit, doesn't seem that useful in my case, although might help someone else I guess.

lukemurray:
Related to this but more to method fields as query field using `[GraphQLField]` I wanted to get some thoughts...

Consider added these to the `Movie` type in the Demo project

```c#
        [GraphQLField]
        public uint DirectorAgeAtRelease => (uint)((Released - Director.Dob).Days / 365);

        [GraphQLField]
        public uint[] AgesOfActorsAtRelease()
        {
            var ages = new List<uint>();
            foreach (var actor in Actors)
            {
                ages.Add((uint)((Released - actor.Person.Dob).Days / 365));
            }
            return ages.ToArray();
        }
```

By default EF doesn't handle this. The expression built by EntityGraphQL calls the method. EF at least knows this and tries to execute it after it has fetched what it does understand.

Issue is that in `DirectorAgeAtRelease` the `Director` property will be `null` as EF doesn't know to fetch it. In `AgesOfActorsAtRelease` the `Actors` property will be `null` for the same reason.

One solution is [Microsoft.EntityFrameworkCore.Proxies](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Proxies/) to lazy load the properties, which comes with its own tradeoffs.

Another one is [EntityFrameworkCore.Projectables](https://github.com/koenbeuk/EntityFrameworkCore.Projectables) which will work for `DirectorAgeAtRelease` by not for `AgesOfActorsAtRelease`. At least it will give a compiler error if you try to add `[Projectable]` to `DirectorAgeAtRelease`. But if you leave it off, and you use EF it will fail.

I feel like including [EntityFrameworkCore.Projectables](https://github.com/koenbeuk/EntityFrameworkCore.Projectables) in the library for this would be useful, but then this library probably shouldn't be pushing the user down a single path. I.e. if they want to use lazy loading.

So that leaves me with somehow making it really clear in docs etc the limitation of marking a method with `[GraphQLField] when using EF and the options around that.

I think the last point is the right direction, but just wanted to write out my thought and get some other thoughts on it, if you have any @bzbetty?

bzbetty:
We use both EF.Projectables, and lazy loading a few places (although I'm not a fan of lazy loading).

I wouldn't include either by default in the library, but I think calling them out in documentation and/examples makes a lot of sense.

PR #249, opened 2022-09-16
replicate and fix #248

### LINKED ISSUE #248: @include not working on single fields
```
query {
    people @include(if: false) {
        id
        name 
    }
}
```

works

```
query {
    person(id:99) @include(if: false) {
        id
        name 
    }
}
```

doesn't

bzbetty:


bzbetty:
Replicates issue #248

PR #257, opened 2022-09-29
Document HTTP request using async

breyed:
Changes the docs to use `ExecuteRequestAsync` instead of `ExecuteRequest`. This not only provides a performance improvement, it also works around an apparent bug. In a production app, we were experiencing extremely high CPU when using the synchronous version of `ExecuteRequest`, rendering the web server unusable. The problem didn't always occur; it seemed to be related to the frequency of GraphQL queries, with successive calls getting progressively slower while always consuming an entire core. Switching to async solved the problem. Switching to async resolved the problem.

The high CPU issue doesn't occur when just calling `ExecuteRequest` directly. I tried adding high frequency calls to a unit test in QueryTests.cs, and they worked just fine. The problem appears to be related to interoperation with ASP.NET Core or Entity Framework Core.

### DIFF
--- a/docs/docs/authorization.md
+++ b/docs/docs/authorization.md
@@ -15,7 +15,7 @@ _Note if you are using the `AddGraphQLSchema()` extension in `EntityGraphQL.AspN
 ```cs
 // Assuming you're in a ASP.NET controller
 // this.User is the current ClaimsPrincipal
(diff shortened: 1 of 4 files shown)


lukemurray:
Wow that is interesting. We have not seen that... although we do not yet have 4.0.0 running in production yet.

PR #259, opened 2022-09-29
Fixes nullable lists

bzbetty:
main change is to make the following correct, although the changes got a bit wider than i'd like and I'm sure there's still issues with the IsNullable method assuming it should look at the first or last byte (unsure if you've ever read on how this byte array works)

```
    //public IEnumerable<Test> Tests { get; set; } = new List<Test>();
            Assert.Contains("tests: [Test!]!", schemaString);
            //public IEnumerable<Test>? Tests2 { get; set; }
            Assert.Contains("tests2: [Test!]\r\n", schemaString);
            //public IEnumerable<Test?> Tests3 { get; set; } = new List<Test>();
            Assert.Contains("tests3: [Test]!", schemaString);
            //public IEnumerable<Test?>? Tests4 { get; set; }
            Assert.Contains("tests4: [Test]\r\n", schemaString);
```

### DIFF
--- a/src/EntityGraphQL/Extensions/Nullability/NullabilityInfo.cs
+++ b/src/EntityGraphQL/Extensions/Nullability/NullabilityInfo.cs
@@ -1,0 +1,71 @@
+﻿#nullable enable
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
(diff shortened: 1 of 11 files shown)


bzbetty:
Pulled in a MIT piece of code from .NET Foundation that provides Nullable Info. makes it much easier to figure out what's nullable and what's not.

PR #260, opened 2022-10-02
support default values in c# methods

bzbetty:
believe this makes default values in mutaitons work

eg

```

        [GraphQLMutation]

        public int DefaultValueTest(int valueWithDefault = 8)
        {
            return valueWithDefault;
        }
```

### DIFF
--- a/src/EntityGraphQL/Schema/SchemaGenerator.cs
+++ b/src/EntityGraphQL/Schema/SchemaGenerator.cs
@@ -1,7 +1,10 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using EntityGraphQL.Directives;
 using EntityGraphQL.Schema.Directives;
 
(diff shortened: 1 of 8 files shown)


bzbetty:
Think this is failing due to another merge. Gets the same Reuse slot error (which seem familiar to me, something about trying to add an Expression type to graphql?)

lukemurray:
Can we add a test/check that the schema/introspection is generating the default values too?

bzbetty:
added in tests/code for sdl/introspection default values

bzbetty:
added a couple more SDL default values checks which requires some form of serialisation, used a json serialiser but it's not quite right, so need another method.

eg
`            schema.Query().AddField("user", new { user = new UserInputObject() { Name = "Steve" } }, (ctx, param) => ctx.Users.FirstOrDefault(), "Return a user by ID");
`
should become

`            Assert.Contains("user(user: userInput = { name:\"Steve\"}): User", schema.ToGraphQLSchemaString());
`
but is currently

`            Assert.Contains("user(user: userInput = { \"name\":\"Steve\"}): User", schema.ToGraphQLSchemaString());
`
which isn't valid SDL.

seems to work for enums, integers, booleans, and empty arrays though

src/EntityGraphQL/Schema/SchemaGenerator.cs
```diff
@@ -1,187 +1,206 @@
...
-                if (!string.IsNullOrEmpty(field.Description))
-                    sb.AppendLine($"\t\"\"\"{EscapeString(field.Description)}\"\"\"");
-                sb.AppendLine($"\t{schema.SchemaFieldNamer(field.Name)}{GetGqlArgs(schema, field)}: {field.ReturnType.GqlTypeForReturnOrArgument}{GetDirectives(field.Directives)}");
-            }
-            sb.AppendLine("}");
-
-            return sb.ToString();
-        }
-    }
+using System;
```
lukemurray, line 1: Why does it show this file 100% changed? Makes it hard to see what did change!

Any highlights?

bzbetty: curious, i must have accidentally formatted it differently. will see if I can clean it up.

there should only be about 2 lines different where it serializes the default value, that said i need to write a custom serializer i think.

bzbetty: image

and GetArgDefaultValue method is entirely new

lukemurray:
My most hated test too :P Need to look at changing it some how..

Thanks though

PR #262, opened 2022-10-04
Do not leak exception messages in 'errors'

rjrudman:
This change stops leaking exceptions into the 'errors' field on a result. This has been raised here: https://github.com/EntityGraphQL/EntityGraphQL/issues/205  

When running in development (read via `IWebHostEnvironment.IsEnvironment("Development")` or when manually creating `SchemaProvider`), messages of exceptions will not be dumped out into the 'errors' field of a query result, unless they implement the newly created (and empty) interface `IExposableException`.

Part of this change was to modify where we're catching and unwrapping exceptions (for example, see [src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs](https://github.com/EntityGraphQL/EntityGraphQL/compare/master...rjrudman:EntityGraphQL:feature/205?expand=1#diff-2649053fc8256fcc5e8681e63a4eeb6c04ce96d96b04b83cafcee6cb7e18b34fL94)) as we were previously immediately rendering the exception message, including that of inner exceptions which should be hidden.

The bulk of the change is [here](https://github.com/EntityGraphQL/EntityGraphQL/compare/master...rjrudman:EntityGraphQL:feature/205?expand=1#diff-35b1d9d391614be9a7d3221dc28bc7c6db245b5a4483da2544b880c2a840a88bL236) which handles the unwrapping and possible obfuscation of thrown errors.

### DIFF
--- a/src/EntityGraphQL.AspNet/Extensions/EntityGraphQLAspNetServiceCollectionExtensions.cs
+++ b/src/EntityGraphQL.AspNet/Extensions/EntityGraphQLAspNetServiceCollectionExtensions.cs
@@ -1,8 +1,10 @@
 using System;
 using EntityGraphQL.Schema;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Hosting;
 
 namespace EntityGraphQL.AspNet
 {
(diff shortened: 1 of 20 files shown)


breyed (commented):
In production, an aggregate exception can result in:
{
  { "Error occurred" },
  { "Error occurred" },
  { "Some error that allows exposure" },
  { "Error occurred" }
}

It would be good to consolidate the generic error messages.

lukemurray:
Good catch. Thanks.

Issue #264, opened 2022-10-05
Union error: "variable '....' of type '....' referenced from scope '', but it is not defined"


Bosch-Eli-Black:
We have some code that uses unions, like so:

```c#
public interface IUser { }

public class Employee : IUser
{
    public string Name { get; set; }
}

public class UsersContext : IUsersContext
{
    public List<IUser> Users { get; set; } = new()
    {
        new Employee
        {
            Name = "test user"
        }
    };
}

public interface IUsersContext
{
    public List<IUser> Users { get; }
}

public class Program
{
    public static void Main(string[] args)
    {
        var builder = WebApplication.CreateBuilder(args);

        builder.Services.AddCors();
        builder.Services.AddTransient<IUsersContext, UsersContext>();
        builder.Services.AddGraphQLSchema<IUsersContext>(options =>
        {
            options.AutoBuildSchemaFromContext = false;

            options.ConfigureSchema = (schema) =>
            {
                schema.AddUnion<IUser>("user", null);
                schema.Type<IUser>().AddAllPossibleTypes();

                schema.UpdateQuery(queryType =>
                {
                    queryType.ReplaceField(
                        "users",
                        db => db.Users,
                        null
                    );
                });
            };
        });

        var app = builder.Build();
        app.UseHttpsRedirection();
        app.UseRouting();
        app.UseCors(builder =>
        {
            builder
            .WithOrigins("*")
            .WithMethods("POST")
            .AllowAnyHeader();
        });
        app.UseEndpoints(endpoints => endpoints.MapGraphQL<IUsersContext>());
        app.Run();
    }
}
```

When I execute this query:
```
{
   users
   {
      ... on Employee
      {
         name
      }
   }
}
```

I get this exception:
```
System.InvalidOperationException
  HResult=0x80131509
  Message=variable 'p_Employee' of type 'Example.Employee' referenced from scope '', but it is not defined
  Source=System.Linq.Expressions
  StackTrace:
   at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
   at System.Linq.Expressions.ParameterExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.Compiler.VariableBinder.Visit(Expression node)
   at System.Linq.Expressions.MemberExpression.Accept(ExpressionVisitor visitor)
[...]

PR #269, opened 2022-10-10
support for serializing subclasses (#264)

bzbetty:
as per #264 .net JSonSerializer doesnt' handle subclasses very well.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-6-0


lukemurray:
Thanks. Can we also update the docs to try and make users aware? Similar to the Configuring System.Text.Json section.

Issue #252, opened 2022-09-16
[GraphQLField]


bzbetty:
split from #241 
> I have plans for a [GraphQLField] to allow a different way of making fields with arguments etc. So perhaps as part of that need to make sure it works well with some of the other libraries.

bzbetty:
Currently trying this in my schema builder manually.  Doesn't yet quite work due to default values (see PR #260)

Bit of a hack but seems that MethodField works for queries.

```
 //Add any method marked as [GraphQLField] as a field
            foreach (var schemaType in schemaProvider.GetNonContextTypes())
            {
                if (schemaType.GqlType == GqlTypeEnum.Object)
                {
                    foreach (var method in schemaType.TypeDotnet.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public))
                    {
                        var graphQlFieldAttribute = method.GetCustomAttribute<GraphQLFieldAttribute>();
                        if (graphQlFieldAttribute == null)
                            continue;

                        var isAsync = method.GetCustomAttribute(typeof(AsyncStateMachineAttribute)) != null || (method.ReturnType.IsGenericType && method.ReturnType.GetGenericTypeDefinition() == typeof(Task<>));
                        var methodAuth = schemaType.Schema.AuthorizationService.GetRequiredAuthFromMember(method);
                        var requiredClaims = methodAuth;
                       
                        if (schemaType.RequiredAuthorization != null)
                            requiredClaims = requiredClaims.Concat(schemaType.RequiredAuthorization);

                        var actualReturnType = GetTypeFromMethodReturn(method.ReturnType, isAsync);
                        var nonListReturnType = actualReturnType.GetNonNullableOrEnumerableType();
                        if (!schemaType.Schema.HasType(nonListReturnType) && options.AutoCreateNewComplexTypes)
                        {
                            schemaType.Schema.AddType(nonListReturnType, nonListReturnType.Name, string.Empty).AddAllFields();                            
                        }
                        var typeName = schemaType.Schema.GetSchemaType(nonListReturnType, null).Name;
                        
                        var returnType = new GqlTypeInfo(() => schemaType.Schema.Type(typeName), actualReturnType, method.IsNullable());
                        var field = new MethodField(schemaType.Schema, schemaType, graphQlFieldAttribute.FieldName, returnType, method, string.Empty, requiredClaims, isAsync, schemaType.Schema.SchemaFieldNamer, options);

[...]

PR #272, opened 2022-10-13
GraphQLFieldAttribute on method PoC

bzbetty:
not working yet, starting a PoC on how this could work


bzbetty:
#252

bzbetty:
seems to work quite well in tandem with EF Projectables

bzbetty:
noticed a couple of small differences in implementation, unsure if they're a big deal

* i had to graphqlignore methods on some of my classes (honestly can't rememer why, doesn't seem to be valid or even an issue in 5.0)
* My GraphQLField methods ended up triggering MakeFieldWithIdArgumentIfExists still, not sure if i made use of that, just diffing the schema atm.

lukemurray:
Hmm, wasn't sure it made sense to still try to build `MakeFieldWithIdArgumentIfExists`. 🤔  So add a require `id` argument to the existing arguments and add a `.Where()` after the method call?

bzbetty:
yeah good point, probably doesn't make sense.  I think i always used it with an ef projectable so it was really just a expression in disguise.  Not a big deal I can add in extra methods if i need them.

bzbetty:
did the 5.0 branch ever get anywhere?

lukemurray:
It did. There was a few edge cases I need to fix up. Just been busy with other things.

PR #274, opened 2022-10-19
bugfix: stop double sql query on base type (4.1 regression)

bzbetty:
So it appears the Expression.TypeIs strategy for interfaces/unions causes entityframework to perform extra queries when AsSplitQuery is enabled

I made a tweak in 4.1 to fix some cases around querying unions without querying multiple types, this meant that when used in conjunction with AsSplitQuery it queried the database once for each type and twice for the base type.

This PR fixes the twice for the base type issue (but not for the once for each type issue).

### DIFF
--- a/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
+++ b/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
@@ -452,9 +452,13 @@ namespace EntityGraphQL.Compiler.Util
                     throw new EntityGraphQLCompilerException("Could not create dynamic type");
 
                 //create a cascading TypeIs X query for each type in query
-                Expression? previous = CreateNewExpression(field.Name, fieldsOnBaseType, out Type _, parentType: baseDynamicType) ?? Expression.Constant(null, baseDynamicType);
+                Expression? previous = CreateNewExpression(field.Name, fieldsOnBaseType, baseDynamicType) ?? Expression.Constant(null, baseDynamicType);
+
                 foreach (var type in validTypes)
                 {
+                    if (type == currentContextParam.Type)
+                        continue;
+
                     var fieldsOnType = fieldExpressions.Values
                        .Where(i => RootType(i.Expression) == null || RootType(i.Expression)!.IsAssignableFrom(type) || typeof(ISchemaType).IsAssignableFrom(RootType(i.Expression)))
                        .ToDictionary(i => i.Field.Name, i => i.Expression);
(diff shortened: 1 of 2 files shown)


lukemurray:
Thanks, do you have an example of the expression it builds (as if you would write it in C#)?

bzbetty:
Given a set of classes

```
public class Job {
    public Person SalesPerson { get; set; }
}

public ReworkJob : Job {
   public int ReworkCategoryId { get; set; }
}
```

with the query
```

query Jobs {
  jobs {
     id
    ... on ReworkJob {
			reworkCategoryId
    }
    salesPerson {
      name
    }
  }
}


```

the following select is now generated in 4.1

```
Select(
            p_Job => (p_Job is ReworkJob)
                ? new Dynamic_jobs_8f86d103-bdbd-43c0-b6b5-2bec9a2a0b99
                {
                    reworkCategoryId = ((ReworkJob)p_Job).ReworkCategoryId,
                    id = ((ReworkJob)p_Job).Id,
                    salesPerson = (((ReworkJob)p_Job).SalesPerson == null)
                        ? null
                        : new Dynamic_salesPerson_0c34301e-fdfe-4fda-8ee4-41a7b0694cbf
                        {
                            name = ((ReworkJob)p_Job).SalesPerson.Name
                        }
                }
                : (p_Job is Job)
                    ? new Dynamic_jobs_87ad6127-dda6-4f3a-8f35-d64f5970d24a
                    {
                        id = ((Job)p_Job).Id,
                        salesPerson = (((Job)p_Job).SalesPerson == null)
                            ? null
                            : new Dynamic_salesPerson_0c34301e-fdfe-4fda-8ee4-41a7b0694cbf
                            {
                                name = ((Job)p_Job).SalesPerson.Name
                            }
                    }
                    : new Dynamic_jobs_87ad6127-dda6-4f3a-8f35-d64f5970d24a
                    {
                        id = p_Job.Id,
                        salesPerson = (p_Job.SalesPerson == null)
                            ? null
                            : new Dynamic_salesPerson_0c34301e-fdfe-4fda-8ee4-41a7b0694cbf
                            {
                                name = p_Job.SalesPerson.Name
                            }
                    })
        .ToList()

```

The PR removes the middle part due to line 459 on [ExpressionUtil.cs].


As i mentioned it's not a huge deal unless you're also using AsSplitQuery() as the sql it generates ends up pulling the salesperson lookup into another query which it repeats 3 times (this PR would reduce it to two).

Issue #275, opened 2022-10-21
ParameterReplacer performance


bzbetty:
So doing some performance profiling recently we've noticed that our issues weren't SQL times but related to the ParameterReplacer class.

Now I'm not saying this class is overly slow - we're passing in a reasonably big query, but we have noticed that each .When or .WhenWhere appened to the query makes a larger and larger impact than the last one.  To the point where one call to ParameterReplacer.Replace is taking over 900ms.

```
Call to Replace 897.8ms
Expression = ctx.Set().AsSplitQuery().IncludeDeletedWhen(ctx, (Convert(arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.filterType, Nullable`1) == Convert(Lost, Nullable`1))).WhereWhen(i => arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.jobTypes.Contains(i.TypeId), (arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.jobTypes != null)).WhereWhen(i => arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.stages.Contains(i.StageId), (arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.stages != null)).WhereWhen(i => (Convert(i.FactoryId, Nullable`1) == arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.factoryId), arg_Dynamic_mergedArgs_96aba847-67ac-4181-892e-99d616252ae3.factoryId.HasValue).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).Where(i => True).OrderBy(Param_0 => Param_0.ProdOrder) 
```

(yes there's a lotof .Where(i => True) in there, i had actual where whens in there before but noticed it didn't seem to matter in terms of performance, each .Where(i => True) added a significant impact.)

No specific method seems to take a lot of time it's just visiting a lot of things.

Thoughts
* Even a tiny fix to the methods on this class would have a large benefit since it seems to go quite deep
* Expressions themselves don't seem to be cached at all (just the GraphQL doc) maybe there's something we could do there

Gonna dig in to this deeper.

lukemurray:
Interesting! Thanks for the heads up. Certainly something I want to look into

[...]

PR #277, opened 2022-10-21
bugfix: ParameterReplacerBenchmark & Performance Fix

bzbetty:
another for #275


lukemurray:
Oh I see you have a more realistic test in `SetupLargerSetOfWhereWhens`. 👍

bzbetty:
> play with more to see if it impacts cached queries

I didn't think there was any expression tree caching? It was something I was thinking of suggesting

bzbetty:
Also have a feeling the parameter replacer runs twice in a row on the same expression so I think there's more room for improvement.

lukemurray:
Yeh, not expression tree caching. But the nodes get cached for matching queries. just need to double check the lifetime of the replacer.

Any ideas where it might be running twice in a row? I'm looking into this general issue now.

bzbetty:
>Any ideas where it might be running twice in a row? I'm looking into this general issue now.

On review I think i was mistaken.

Field.PrepareFieldExpression called replace, which then returns to GetExpression which called it too, but looks like the arguments are different.  Could potentially have savings by traversing once instead of twice, but wasn't as wasteful as I first thought.

PR #278, opened 2022-10-30
feat: optionalfield arguments - (null, undefined or value)

bzbetty:
Sometimes have the usecase where I want something akin to undefined in javascript that I can treat differently to a null value - for example a nullable date in the database that I don't want to overwrite in a mutation if it's considered undefined.

needs more tests and documentation, but any thought on the direction?

### DIFF
--- a/src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
+++ b/src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
@@ -6,6 +6,7 @@ using EntityGraphQL.Compiler.Util;
 using System;
 using EntityGraphQL.Extensions;
 using HotChocolate.Language;
+using Nullability;
 
 namespace EntityGraphQL.Compiler
 {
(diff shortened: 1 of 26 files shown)


bzbetty:
hmm not sure this is actually possible - as the OptionalField type doesn't make it through to docVars so it always falls back to a default value of null.

lukemurray:
What you're trying here is something like this?

```cs
public class MyArgs {
  public OptionalField<int> SomeInt { get; set; }
}
```

And with JSON variables passed

```js
// Sets property to 1
{
    "someInt": 1
}

// Sets property to null
{
    "someInt": null
}

// Sets property to "undefined" / not set
{
}
```

bzbetty:
Yes exactly like that.

I've managed to get a PoC working using https://github.com/Foritus/optional-dot-net after adding implicit casting #285 so while it might still be worth trying to build into EQL, I don't have an immediate need for it.

lukemurray:
Thanks. I might add it to the list. We use a more verbose pattern where we add a `clearX` argument.

PR #280, opened 2022-10-31
Fix for mutation when they return interfaces/unions

bzbetty:
mutations were running the wrong code if they returned an interface/union type


lukemurray:
Can you add a test that previously would fail and now passes so we don't undo this accidentally in the future? Thanks

bzbetty:
added a test that behaves differently with the fix but can't pass without #279

bzbetty:
interestingly enough the issue only appears to be when returning single entities not list from mutations, which I guess might be a clue to fix this one in a better way.

bzbetty:
updated to have a test that passes now

PR #281, opened 2022-11-03
Implement whitelisted exceptions

rjrudman:
Related to #270  

This change allows consumers of the library to specify whitelisted exceptions on top of the provided `IExposableException`.

This is required because users may not be able to modify the exceptions being thrown in order to have them implement `IExposableException`.

Exceptions may be whitelisted with an exact type match, or whitelisted including all types that inherit it.

### DIFF
--- a/src/EntityGraphQL/Schema/SchemaProvider.cs
+++ b/src/EntityGraphQL/Schema/SchemaProvider.cs
@@ -45,7 +45,7 @@ namespace EntityGraphQL.Schema
 
         public IDictionary<Type, ICustomTypeConverter> TypeConverters { get; } = new Dictionary<Type, ICustomTypeConverter>();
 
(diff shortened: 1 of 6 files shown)


src/EntityGraphQL/Schema/SchemaProvider.cs
```diff
@@ -45,6 +45,8 @@ public class SchemaProvider<TContextType> : ISchemaProvider
 
         public IDictionary<Type, ICustomTypeConverter> TypeConverters { get; } = new Dictionary<Type, ICustomTypeConverter>();
 
+        public List<WhitelistedException> WhitelistedException { get; } = new();
```
lukemurray, line 48: As we use `IsAllowed` below can we name this `AllowedExceptions`?

rjrudman: Yep, done!

Issue #221, opened 2022-08-25
Extracting data from ICollection, List, etc. throws exception when null


Bosch-Eli-Black:
(Originally posted at https://github.com/EntityGraphQL/EntityGraphQL/issues/204#issuecomment-1223643790)

**Code**

```c#
#nullable enable
public class UserDbContext : DbContext {
   public DbSet<UserWrapper> UserWrappers { get; set; }
}

public class UserWrapper {
   public string Json { get; set; }
}

public class User {
    public ICollection<Role>? Roles { get; set; } = null;
}

services.AddGraphQLSchema<UserDbContext>(options =>
{
    options.ConfigureSchema = ConfigureSchema;
});

public static void ConfigureSchema(SchemaProvider<UserDbContext> demoSchema)
{
    demoSchema.AddType<User>("user", null).AddAllFields();

    // Add custom root fields
    demoSchema.UpdateQuery(queryType =>
    {
        queryType.ReplaceField(
            "users",
            new
            {
                ...
            },
            (db, args) => db.UserWrappers.Select(userWrapper => JsonConvert.DeserializeObject<User>(userWrapper.Json)).ToList(),
            "List of users"
        );
    });
}
```

**Query**
```
{"Query":" { users { roles { id } }"}
```

**Error**
```json
{"errors":[{"message":"Field 'users' - Value cannot be null. (Parameter 'source')"}]}
```

lukemurray:
The built expressions do null check when there are service fields involved. As it kind of makes an assumption that without service fields the expression will be handled by EF. In this case you are using EF.

I think EF is choosing to run some of the expression in memory after fetching the data it can. Then either `JsonConvert.DeserializeObject<User>(userWrapper.Json)` returns `null` for an item or the deserialised item has a `null` `Roles` property.

Makes me think of two points
1. this should be better highlighted in the docs
2. there likely should be an option to have this happen regardless of service fields

Interesting you don't "need" a service. But, if you could change the field to use a service (like make a deserialiser service) it may work with the current release.

Interesting use case though that needs more thought. I originally thought it was going to be something simple!

Bosch-Eli-Black:
@bzbetty Thanks for the fix 🙂  Do you agree that it's best to silently return nothing? I think that's probably the desired behaviour here, but I could also see an argument for having it throw.

bzbetty:
[...]

PR #291, opened 2022-11-21
ResolveWithService being called twice due to Expression.Condition 

bzbetty:
the fix for #221 added a expression.condition, this accidentally evaluates ResolveWithService calls twice which can make multiple db requests.

Using Expression.Coalesce is a better choice but does mean we need to return an empty list instead of null (don't think that's a bad thing but it is a change).


lukemurray:
Why can't we return `null`? It'd be good to respect the schema and if `null` is a valid response it should be used.

Might need to use the wrapInNullCheck logic here too.

bzbetty:
i wouldn't have an issue with a null return, i believe it was Bosch-Eli-Black that had the original issue.

lukemurray:
I mean why does using `Expression.Coalesce` force us to return a `null`?

I guess I'll dig into it :)

bzbetty:
ah right, can't return null because of the ToList call that happens afterwards.

PR #301, opened 2023-03-18
Changes to Method Argument Reflection

lukemurray:
Previously if `AutoCreateInputTypes` was enabled we didn't know if a parameter should be a GraphQL argument or an injected service unless you used `[GraphQLArguments]`. But this meant you couldn't have complex types as parameters in the method and have them reflected in the schema (`[GraphQLArguments]` flattens the arguments in the schema). This has been refactored to be predictable. 

`AutoCreateInputTypes` now defaults to `true` and you will have to add some attributes to your parameters or classes.

`[GraphQLInputType]` will include the parameter as an argument and use the type as an input type. `[GraphQLArguments]` will flatten the properties of that parameter type into many arguments in the schema.

When looking for a methods parameters, EntityGraphQL will

1. First all scalar / non-complex types will be added at arguments in the schema.

2. If parameter type or enum type is already in the schema it will be added at an argument.

2. Any argument or type with `GraphQLInputTypeAttribute` or `GraphQLArgumentsAttribute` found will be added as schema arguments.

3. If no attributes are found it will assume they are services and not add them to the schema. *I.e. Label your arguments with the attributes or add them to the schema beforehand.*

`AutoCreateInputTypes` now only controls if the type of the argument should be added to the schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `SchemaBuilderMethodOptions` removed, see updated properties on `SchemaBuilderOptions` and upgrade docs. This was because you can also now add methods ad query fields with `GraphQLFieldAttribute`
 - `SchemaBuilderOptions.AutoCreateInputTypes` now defaults to `true`. Meaning in `SchemaBuilder` when adding mutations etc any complex types will be added to the schema if they are not there already.
 - The rules for reflection on method parameters have been changed to make them clearer. Teh the upgrade to 5.0 docs and the mutation docs that cover examples.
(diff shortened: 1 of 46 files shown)


lukemurray:
@bzbetty Keen to can some other eyes on this (anyone else that reads this too).

I've tried to clean up the reflection rules of methods (this will be used in mutations, subscriptions and the new `[GraphQLField]` query fields).

Also see the update mutation docs which contain examples. Hopefully they make sense, but basically
> Label your arguments with the `GraphQLInputTypeAttribute` or `GraphQLArgumentsAttribute` or  add them to the schema beforehand


Thoughts, feedback?

bzbetty:
Did the AddNonAttributedMethods option dissapear?

Issue #304, opened 2023-05-03
[Feature] Add the ability to secure the GraphQL Endpoint in ASP.net Core


RubberChickenParadise:
Currently, due to the use of mapping endpoints, there is not an easy way to secure the entire graphql endpoint.  The object level security works but in cases where only authenticated users should have access to any data, even the schema, there is no easy way to accomplish this.

Adding the ability to apply an Authorization Policy to the endpoint itself would be very helpful for the security needs of several applications.

lukemurray:
merged coming shortly in the 5.0 release

PR #305, opened 2023-05-03
Add the ability to use RequireAuthorization when using MapPost to add authentication policies to the graphql endpoint

RubberChickenParadise:
#304 Add the ability to pass authorization policies that should be applied to the Post Endpoint.  By passing in the authorization policies and using RequireAuthorization, the entire graphql endpoint can be easily secured with an authorization policy.

I added an additional optional string array parameter to the `MapGraphQL` extension and passed it to the `RequireAuthorization()` method for the `MapPost` call.

This allows using .net asp.net core authorization policies on the endpoint to enable ensuring users are authenticated

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -112,6 +112,37 @@ _You can authorize the route how ever you wish using ASP.NET. See the Authorizat
 
 You can also expose any endpoint over any protocol you'd like. We'll use HTTP/S for these examples.
 
+## Securing the route in ASP.NET core
+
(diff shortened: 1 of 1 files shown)


lukemurray:
Thanks for this! Any chance you could update the docs with an example of how to use please?

RubberChickenParadise:
I will work on getting the documentation updated if the coding changes will be accepted with the updated documentation.

Thanks
Jessica
Get Outlook for iOS<https://aka.ms/o0ukef>

lukemurray:
Yes, the code changes look good. Thank you!

RubberChickenParadise:
@lukemurray Documentation updated.  I put it in the Quick Start guide since security on the endpoint should be a universal requirement for most applications that want to restrict access.

PR #325, opened 2023-10-28
first pass at a Aggregate extension

lukemurray:
Add a extension to easily add aggregation fields for a list field


lukemurray:
added in 6-beta8

PR #333, opened 2023-12-06
Working on fixes to method and service calls on interfaces/unions

bzbetty:
WIP - have some broken tests, haven't gotten head around how to fix.

### DIFF
--- a/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
+++ b/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
@@ -226,6 +226,10 @@ namespace EntityGraphQL.Compiler
                     // we selected ctx.SomeField on the first execution and on the second execution we use newCtx.ctx_SomeField
                     // if ParentNode?.HasServices == true the above has been done and we just need to replace the 
                     // expression, not rebuild it with a different name
+
+
+                    //TODO replacementNextFieldContext is currently of the wrong type, needs to be converted
+
                     var expReplacer = new ExpressionReplacer(expressionsToReplace, replacementNextFieldContext, ParentNode?.HasServices == true, isRoot && HasServices);
                     nextFieldContext = expReplacer.Replace(nextFieldContext!);
                 }
--- a/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
+++ b/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
@@ -156,7 +156,7 @@ namespace EntityGraphQL.Compiler
             if (withoutServiceFields && Field?.ExtractedFieldsFromServices != null)
                 return Field.ExtractedFieldsFromServices.ToList();
 
-            return withoutServiceFields && HasServices ? new List<BaseGraphQLField>() : new List<BaseGraphQLField> { field ?? this };
+            return withoutServiceFields && HasServices ? [] : new List<BaseGraphQLField> { field ?? this };
         }
 
         public void AddField(BaseGraphQLField field)
--- a/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
+++ b/src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
@@ -114,20 +114,20 @@ namespace EntityGraphQL.Compiler
         /// <param name="contextChanged">If true the context has changed. This means we are compiling/executing against the result ofa pre-selection without service fields</param>
         /// <param name="replacer">Replace used to make changes to expressions</param>
         /// <returns></returns>
-        public Expression? GetNodeExpression(CompileContext compileContext, IServiceProvider? serviceProvider, List<GraphQLFragmentStatement> fragments, ParameterExpression? docParam, object? docVariables, ParameterExpression schemaContext, bool withoutServiceFields, Expression? replacementNextFieldContext, bool isRoot, bool contextChanged, ParameterReplacer replacer)
+        public Expression? GetNodeExpression(CompileContext compileContext, IServiceProvider? serviceProvider, List<GraphQLFragmentStatement> fragments, ParameterExpression? docParam, object? docVariables, ParameterExpression schemaContext, bool withoutServiceFields, Expression? replacementNextFieldContext, List<Type>? possibleNextContextTypes, bool isRoot, bool contextChanged, ParameterReplacer replacer)
         {
             IGraphQLNode? fieldNode = ProcessDirectivesVisitNode(LocationForDirectives, this, docParam, docVariables);
 
             if (fieldNode == null)
                 return null;
 
-            return ((BaseGraphQLField)fieldNode).GetFieldExpression(compileContext, serviceProvider, fragments, docParam, docVariables, schemaContext, withoutServiceFields, replacementNextFieldContext, isRoot, contextChanged, replacer);
+            return ((BaseGraphQLField)fieldNode).GetFieldExpression(compileContext, serviceProvider, fragments, docParam, docVariables, schemaContext, withoutServiceFields, replacementNextFieldContext, possibleNextContextTypes, isRoot, contextChanged, replacer);
         }
 
         /// <summary>
         /// GetNodeExpression but without the directive processing as we do not want to process continuously
         /// </summary>
-        protected abstract Expression? GetFieldExpression(CompileContext compileContext, IServiceProvider? serviceProvider, List<GraphQLFragmentStatement> fragments, ParameterExpression? docParam, object? docVariables, ParameterExpression schemaContext, bool withoutServiceFields, Expression? replacementNextFieldContext, bool isRoot, bool contextChanged, ParameterReplacer replacer);
+        protected abstract Expression? GetFieldExpression(CompileContext compileContext, IServiceProvider? serviceProvider, List<GraphQLFragmentStatement> fragments, ParameterExpression? docParam, object? docVariables, ParameterExpression schemaContext, bool withoutServiceFields, Expression? replacementNextFieldContext, List<Type>? possibleNextContextTypes, bool isRoot, bool contextChanged, ParameterReplacer replacer);
 
         public IEnumerable<BaseGraphQLField> Expand(CompileContext compileContext, List<GraphQLFragmentStatement> fragments, bool withoutServiceFields, Expression fieldContext, ParameterExpression? docParam, object? docVariables)
         {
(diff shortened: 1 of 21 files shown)


bzbetty:
TShirtOrderItem inherits from OrderItem
```
       schemaProvider.UpdateType<TShirtOrderItem>(Order =>
       {
           Order.AddField("statusAsString", "Get the order status as a string")
           .ResolveWithService<TestService>((o, srv) => srv.FormatTShirtPropertiesAsString(o.Colour, o.Size));
       });
```

```
                fragment orderItem on OrderItem {
                    id
                    ... on TShirtOrderItem {
                        statusAsString
                    }
                }
```

leads to  GraphQLExtractedField GetNodeExpression
 System.Exception : Field 'order' - 'o_Colour' is not a member of type 'Dynamic_baseDynamicType_30973570-91d0-418f-b100-741f49c310a4' (Parameter 'propertyOrFieldName')

* seems that it's looking on the base type instead of the inherited type for the property, i think because replacementNextFieldContext is wrong in this case (maybe unions need to be pushed further up the parsing tree, was always a bit of a hack in ExpressionUtil)
* name seems off - eg the "o_" unsure if that's normally in the expression or if it's not being removed for some reason

bzbetty:
* includes a fix around detecting root type 
* removes ConversionVisitor in favour of ParameterReplacer as it covers a lot more cases

bzbetty:
step forward, updated the inline fragment to expand fields.  Now correctly queries the database.

now stuck in the BaseGraphQLField.ReplaceContext trying to do another Convert statement.

bzbetty:
```
{
                ""query"": ""query Order($orderId: Int) {
                    order(id: $orderId) {
                        ...order
                    }
                }

                fragment order on Order {
                    orderItems {
                        ...orderItem
                    }
                }

                fragment orderItem on OrderItem {
                    id
                    ... on TShirtOrderItem {
                        statusAsString
                    }
                }"",
                ""variables"": {
                    ""orderId"": ""1""
                }
            }
```
builds the expression
```


cxt_TestContext.Orders.Where(
    arrcxt_Order => arrcxt_Order.Id == args_Dynamic_orderItems_ae2fa559-4250-41b0-aac1-2ae0509e42fb_exec.id.Value).Select(
    p_Order => new Dynamic_order_d9d63083-c301-44f3-ae9a-71b7ae91ecad
    {
        orderItems = p_Order.OrderItems.Select(
            p_OrderItem => (p_OrderItem is InheritanceAdvancedTest.TShirtOrderItem)
                ? new Dynamic_orderItems_71035a4c-ffe4-4999-972a-ba9318e43598
                {
                    o_Colour = ((InheritanceAdvancedTest.TShirtOrderItem)p_OrderItem).Colour,
                    o_Size = ((InheritanceAdvancedTest.TShirtOrderItem)p_OrderItem).Size,
                    id = p_OrderItem.Id
                }
                : new Dynamic_orderItemsbaseDynamicType_9cefd2a7-c1ae-416a-998b-ba7ebfc88450
                {
                    id = p_OrderItem.Id
                }).ToListWithNullCheck(false)
    })
```

which is great, but when it tries putting the service call back in there by removing

```
         o_Colour = ((InheritanceAdvancedTest.TShirtOrderItem)p_OrderItem).Colour,
          o_Size = ((InheritanceAdvancedTest.TShirtOrderItem)p_OrderItem).Size,
```

it gets ```    System.Exception : Field 'order' - 'o_Colour' is not a member of type 'Dynamic_orderItemsbaseDynamicType_9cefd2a7-c1ae-416a-998b-ba7ebfc88450' (Parameter 'propertyOrFieldName')```

because it's looking it up on the base type instead of Dynamic_orderItems_71035a4c-ffe4-4999-972a-ba9318e43598.

If i had access to that c# type then a cast would be easy, but can't figure out where i would even find it.

bzbetty:
added a terrible hack to make the test pass

bzbetty:
@lukemurray any ideas on how to do this without reflection? struggling to figure out the best way to pass this information around.

lukemurray:
I haven't looked at the details.

But try to do reflection at schema building time and hold the info in the field data etc. Second to that is at parsing/"compile" time and how info in the GraphQL*Fields as they are cacheable.

bzbetty:
>But try to do reflection at schema building time and hold the info in the field data etc.
This is for the dynamic types built at query time, so can't be at building time.

> Second to that is at parsing/"compile" time and how info in the GraphQL*Fields as they are cacheable.
I suspect there's a way to build the expression tree better for inherited types that would make it easier to grab this.  Potentially casting too low down and there'd be a way to make the parameter the right type.

eg there's a lot of casting in this statement
````
.Select(animal => 
     animal is Dog ? new { property = ((Dog)animal).Property,  property2 = ((Dog)animal).Property2 }
     : animal is Cat ? new { property = ((Cat)animal).Property3,  property4 = ((Cat)animal).Property4 }
     : new { property5  = animal.property5 }  
```

wonder if expressions can do soemthing like

````
.Select(animal => 
     animal is Dog d ? new { property = d.Property,  property2 = d.Property2 }
     : animal is Cat c ? new { property = c.Property3,  property4 = c.Property4 }
     : new { property5  = animal.property5 }  
```
unsure that'd fix it.



Otherwise at the very least if there was a way to grab the dynamic types that were built for this query it'd be a lot less classes to check instead of reflection.

lukemurray:
Yeh, Expression need some more love in dotnet.

`animal is Dog d` as far as I know is not doable in expressions today (like many other things, async, ?., etc).

The code generated is pretty fast though, I'm not the the casting would be a real issue?

Sorry, just getting back into this, it seems the test pass? Is there a case will still need to handle?

bzbetty:
> The code generated is pretty fast though, I'm not the the casting would be a real issue?

It's not so much the performance of the casting I'm concerned about, it was the figuring out which dynamic type to create.  Especially when ti comes to unwrapping service calls

> Sorry, just getting back into this, it seems the test pass? Is there a case will still need to handle?

It works, but has that ugly reflection hack at runtime that I can't figure out how to get around.

src/EntityGraphQL/Compiler/GqlNodes/BaseGraphQLField.cs
```diff
@@ -226,6 +226,10 @@ protected Expression ReplaceContext(Expression replacementNextFieldContext, bool
                     // we selected ctx.SomeField on the first execution and on the second execution we use newCtx.ctx_SomeField
                     // if ParentNode?.HasServices == true the above has been done and we just need to replace the 
                     // expression, not rebuild it with a different name
+
+
+                    //TODO replacementNextFieldContext is currently of the wrong type, needs to be converted
```
lukemurray, line 231: Can we have a test that hits this case and fails?

lukemurray:
> > The code generated is pretty fast though, I'm not the the casting would be a real issue?
> 
> It's not so much the performance of the casting I'm concerned about, it was the figuring out which dynamic type to create. Especially when ti comes to unwrapping service calls. I thought casting at a different level might provide enough information to figure that out (or remove the need), but I'm probably barking up the wrong tree
> 
> > Sorry, just getting back into this, it seems the test pass? Is there a case will still need to handle?
> 
> It works, but has that ugly reflection hack at runtime that I can't figure out how to get around.

Ahh I see the "//HACK - works but will be slow" :)

I'll have a better look and think about it this weekend.

lukemurray:
Every time I look at this, I start to get my head around it, but then get busy with other things...

The approach I am taking is trying to bring the sub types up to `GraphQLExtractedField` and loop through them in `GetNodeExpression`

```
if (replacementNextFieldContext.Type.GetProperty(Name) != null || replacementNextFieldContext.Type.GetField(Name) != null)
            {
                // extracted fields get flatten as they are selected in the first pass. The new expression can be built
                return Expression.PropertyOrField(replacementNextFieldContext, Name);
            }
            else if (subTypes != null)
            {
                foreach (var type in subTypes)
                {
                    if (type.GetProperty(Name) != null || type.GetField(Name) != null)
                    {
                        return Expression.PropertyOrField(Expression.Convert(replacementNextFieldContext!, type), Name);
                    }
                }
            }
```

Issue I'm having is figuring out the best place to set the sub types, as the dynamic ones are created much later than the extracted fields are originally extracted. So it likely means more moving around.

bzbetty:
Glad it's not simple :) i bashed my head at this one for many hours before even getting an idea what the issue was, then many more before coming up with the hack.

While it would be nice to have a fix, it's a fairly rare situation, i'm currently in prod with the hack and the performance hasn't been an issue.

my suspicision has always been that the inheritance code I stuck in dealt with it too late in the pipe, that is that during the parsing phase more information probably should have been saved somewhere  which would have made the code in MakeSelectWithDynamicType easier to write (eg the dynamic types being created much earlier and saved on the field would make it accessable where the hack is.

lukemurray:
@bzbetty Here is an example of not using the `GetAssemblies()` call. Pulls up the knows dynamic types to when they are created.

Might still be a better way, but I don't think there is, as a query could have `@skip/@include` directives on a field that would change the required dynamic type. So where it currently sits I _think_ is the best place.

There are probably more edge cases. Be keen to know if this works for you code - tests all pass.

bzbetty:
Certainly seems much better, and should work in the majority of cases.

I think it would be technically possible for multiple types of have a field with the same name (which isn't inherited), which could potentially mess with this. However seems unlikely to happen.

bzbetty:
Found a new case of the bug - same thing but when the service also takes arguments.

pushed a test and a fix using the naive hack

bzbetty:
>also weirdly had to reference the fields in the query too, which makes me think the fix isn't actually correct yet
hmm yes this seems to be as its looking for "colour" where the field is actually "o_colour", so something isn't right.

bzbetty:
hmm actually turns out while it executed it wasn't returning the data correctly.

Added some data checks so the tests now fail again (could have sworn this worked at some point)

Interestingly it seems to be building the expressions wrongly.

**1st expression (with extracted arguments)**

image
s

**2nd expression (should be calling the service?)**

image

2nd expression appears to be wrong, belive it's because the argument is a dynamic type but then the typeis checks expect it to be the original entity types so it finds no fields.

Running as non inherited the queries look like this

image

bzbetty:
thoughts on potential fixes

* Update CreateNewExpression to add an implicit conversion back to the entity (may allow the Expression.TypeIs check to work).
* Replace TypeIs check with something else in the case when there's no overlap between currentContextParam.Type and validTypes
* pass PossibleNextContextTypes into MakeSelectWithDynamicType as a potential way to map between dynamic types
* would creating one huge expression tree and executing it once (with a tolist in the middle) keep more context than two trees? (guessing not, just spitballing)
*

lukemurray:
Sorry, where is the "typeis checks expect it to be the original entity" happening?

lukemurray:
Thanks for the info and tests. I have the first one working. just debugging the second one now. Basically we have the information (maybe not obvious) and have to choose a different path if we are not doing the "EF only first pass"

lukemurray:
Latest push fixes the issues, which likely would have cause other similar issues we haven't caught before.

PR #344, opened 2024-02-15
add bulk resolver for object projection fields under list fields

elliotwannemacher-insource:


lukemurray:
Thanks for this. 

I'll be able to have a look at `TestServicesBulkResolverFullObjectWithPaging`.

I think the issue will be that the `FieldExtension`s will need to implement something like `HandleBulkResolve`

lukemurray:
Bringing this into master now. If you find any other edge cases please make a new PR :). Thanks

Issue #345, opened 2024-02-23
Use OutputCache with Entity GraphQl?


PhilippeWaeber:
I'm reposting a [question asked on Stackoverflow](https://stackoverflow.com/questions/78043723/use-outputcache-with-entity-graphql). I'm not sure if this is related to `EntityGraphQL.AspNet` or asp.net core/output caching itself.

---------------------------------------------

I'm trying to setup [output caching][1] using [Entity GraphQl][2], but can't get it to work. It's a net8.0 project with `EntityGraphQL.AspNet` 5.1.1.

Here's my reduced `Program.cs` 

```c#
using EntityGraphQL.AspNet;
using MedicalData.DataAccess;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers();

builder.Services.AddOutputCache();

builder.Services.AddEndpointsApiExplorer();

builder.Services.AddSwaggerGen();

builder.Services.AddMedicalDataDataAccess(builder.Configuration.GetSection("MedicalDataDataAccess").Bind);
builder.Services.AddGraphQLSchema<MedicalDataDbContext>();

var app = builder.Build();

app.UseHttpsRedirection();

app.UseOutputCache();

app.MapGraphQL<MedicalDataDbContext>(
    configureEndpoint: endpointConventionBuilder =>
        endpointConventionBuilder.CacheOutput(outputCachePolicyBuilder =>
        {
            outputCachePolicyBuilder.VaryByValue(context =>
            {
                context.Request.EnableBuffering();

                using var bodyReader = new StreamReader(context.Request.Body, leaveOpen: true);
                var bodyContent = bodyReader.ReadToEndAsync()
                                            .Result
                                            .Replace(" ", "");

                // Reset the stream position to enable subsequent reads
                context.Request.Body.Position = 0;

                Console.WriteLine($"VaryByValue: {bodyContent}");

                return new KeyValuePair<string, string>("graphQl-requestBody", bodyContent);
            });

            outputCachePolicyBuilder.Expire(TimeSpan.FromHours(6));
        }));
app.MapGraphQLAltair();

app.MapControllers();

app.Run();
```

When executing the same graphql query multiple times, it is never being retrieved from the cache. The vary by value function is being executed each time, then the EF query:

    VaryByValue: {"query":"{\narticle(id:3){\nupdatedOn\ndescription{\ncurrent\n}\n}\n}","variables":{},"operationName":null}
[...]

PR #348, opened 2024-03-06
Add markdown documentation for caching

PhilippeWaeber:
- Describe wiring in Program.cs
- Describe custom IOutputCachePolicy
- Add it as the last page, until we decide where to put it

Relates to #345

### DIFF
--- a/docs/docs/caching.md
+++ b/docs/docs/caching.md
@@ -2,8 +2,11 @@
 sidebar_position: 13
 ---
 
-# Output caching
-The [output caching](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/overview?view=aspnetcore-8.0#output-caching), available in asp.net core, can be used to cache GraphQL requests/responses.
+# Caching
+Several ways of caching data are available in an asp.net project. Some will be described here.
+
+## Output caching
+The [output caching](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/output), available in asp.net core, can be used to cache GraphQL requests/responses.
 
 Register required services (`builder.Services.AddXyz()`) and set up the request pipeline (`app.UseXyz()`) appropriately.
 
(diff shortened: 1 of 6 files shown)


PhilippeWaeber:
@lukemurray I've added the page last but would suggest adding it after validation (index 7) and move the ones below one index further down.
Any other feedback welcome. Feel free to adapt the doc.

PhilippeWaeber:
@bzbetty Could be worth setting up a `Chaching` page with different chapters. Let's wait for input.

PhilippeWaeber:
@bzbetty I've slightly adapted the page to hold documentation for different caching techniques. Please add your part in there.
@lukemurray Where to put it? After Validation (index 7) ? Anything else to do besides setting `sidebar_position: 13` in the file itself?

lukemurray:
That should be it. You can run `npm start` from the `docs` directory to test it.

PhilippeWaeber:
Cheers, looks good to me 👍 
Thanks for your help an this lib.

bzbetty:
Not worth writing a lot about EFCoreSecondLevelCacheInterceptor as it's fairly well documented itself and doesn't require EQL to be changed.  But here's some brief docs, i could put them in another PR once this is merged.



## EFCore Second Leve Cache Interceptor

EFCoreSecondLevelCacheInterceptor is a library that caches the result of EF commands and automatically invalidates cached data by watching what tables are updated via EF.  This makes it a really easy way to speed up the performance of EntityGraphQL without the overhead of manual invalidation.  EQL still needs to parse queries and convert them to expression trees, so there's still more cpu than output caching.

It supports numerous caching methods through its support of EasyCaching.Core (InMemory, Redis, Sql, Disk + more).

For full setup instructions see their documentation on their github page (https://github.com/VahidN/EFCoreSecondLevelCacheInterceptor) however it can be as simple as enabling efsecondlevelcache, adding a caching provider and adding the intercepter to EFCore.

```
  services.AddEFSecondLevelCache(options => {
     options.UseMemoryCacheProvider();
  });

  services.AddDbContextPool<ApplicationDbContext>((serviceProvider, optionsBuilder) => {
     optionsBuilder
        .UseSqlServer(...)
        .AddInterceptors(serviceProvider.GetRequiredService<SecondLevelCacheInterceptor>()));
  });

```

PR #382, opened 2024-06-25
fix #303 - Extract Service Parameters to allow filtering on services

### LINKED ISSUE #303: UseFilter referencing a service field in the filter arg fails
Repro

```cs
[Fact]
        public void TestOffsetPagingWithOthersAndServices()
        {
            var schema = SchemaBuilder.FromObject<TestDataContext>();
            var data = new TestDataContext();
            data.People.Add(new Person { Id = 1, Name = "Jill", LastName = "Frank", Birthday = DateTime.Now.AddYears(22) });
            data.People.Add(new Person { Id = 2, Name = "Cheryl", LastName = "Frank", Birthday = DateTime.Now.AddYears(10) });

            schema.Query().ReplaceField("people", ctx => ctx.People, "Return list of people")
                .UseFilter();
            schema.Type<Person>().AddField("age", "Persons age")
                .ResolveWithService<AgeService>((person, ager) => ager.GetAge(person.Birthday));
            var gql = new QueryRequest
            {
                Query = @"{
                    people(filter: ""age > 21"") {
                        name id age lastName
                    }
                }",
            };

            var serviceCollection = new ServiceCollection();
            var ager = new AgeService();
            serviceCollection.AddSingleton(ager);

            var result = schema.ExecuteRequest(gql, data, serviceCollection.BuildServiceProvider(), null);
            Assert.Null(result.Errors);

            dynamic people = result.Data["people"];
            Assert.Equal(1, Enumerable.Count(people));
            var person1 = Enumerable.ElementAt(people, 0);
            Assert.Equal("Frank", person1.lastName);
            Assert.Equal("Jill", person1.name);
        }
```

crjc:
as mentioned in #303, this fix appears to be working as expected with **EF Core** for using services in the filter expression.

This solution looks at any arguments in the query. If the argument's value is a query (e.g. filter:), then it extracts any ParameterExpressions from this query and adds these as services.

The test from #303 is included here and passes. It fails when ran on master.


lukemurray:
Thanks.

This does break the `ExecutionOptions.ExecuteServiceFieldsSeparately` option as it will try to include the service fields in the expression regardless of that option (when using a filter with one). Ideally to maintain that support it needs to remove those fields from the filter and only include them in the second pass.

It does raise the interesting point that perhaps newer (post 3.1 maybe) EF handles this itself now - or at least better. But I would need to test that a lot more or confirm that somehow.

A lot of the tests skip EF altogether, so I plan to add a new test project to add some of these cases to that at least goes through EF code.

lukemurray:
OK reading [this](https://learn.microsoft.com/en-us/ef/core/querying/client-eval) it seems EF only handles the auto client translation at the top level

>If EF Core detects an expression, in any place other than the top-level projection, which can't be translated to the server, then it throws a runtime exception

This library does it at any level. Although that can come with not so obvious performance impacts, which I should update the docs on.

I'll add more tests for that as well.

crjc:
Thanks for your comments. I now realise I was mostly oblivious to how the second pass worked and the `ExecuteServiceFieldsSeparately` option.

I could have just set `ExecuteServiceFieldsSeparately` to false and allow EF to handle the services for me, instead of attempting a fix which essentially breaks the 2-pass functionality in this library!

I need to do some more testing on my end but I think everything works (for my use case) as expected with this option turned off. 

Do you have an idea of a GraphQL query which would trigger the EF Core runtime exception?

lukemurray:
Based on the docs, any time you use a method or property it doesn't know how to translate into SQL it will not work. However there is a setting to either throw an error in this case, or it fetches more data and executes the non SQL part in memory.

Pretty sure in dotnet 6 the default is to throw the error as we get this if we use things not knows to the DB.

What the library tried to do is minimise over fetching (as we know what we need) from the DB then execute the next bit in memory.

If you have no performance issues with how you are using it, then great.

PR #387, opened 2024-07-15
fix: use AndAlso for queryparser to match behaviour of older versions

crjc:
Example query:
```
{ 
    users (filter: "isActive == true && isActive == false") { 
        name 
    } 
}
```

5.3.0: 
- Worked as expected, returns no results.
- Generated the following expression:
```
p_EFCoreDemoContext.Users
    .Where(q_User => ((q_User.IsActive == True) AndAlso (q_User.IsActive == False)))
    .Select(p_User => new Dynamic_users_...() {name = p_User.Name})}	
```

5.4.0: 
- Pomelo MySQL Provider throws a MySqlException due to an error in SQL syntax
- However, SqlServer provider continues to behave fine
- This lib generates the following expression, uses of `And` instead of `AndAlso`.
```
p_EFCoreDemoContext.Users
    .Where(q_User => ((q_User.IsActive == True) And (q_User.IsActive == False)))
    .Select(p_User => new Dynamic_users_...() {name = p_User.Name})
```

Can we update the parser to use `AndAlso` again please?


lukemurray:
Big miss here by me! Sorry.

PR #405, opened 2024-09-05
Filter extension methods

Jimmy-ZTC:
Use reflection to look for extension methods.  This allows you to add functions to the filter syntax.

I acknowledge that you might have something else in mind to add functions to the filter syntax.  My use case was exposing sqlite's regex function.

```
query {
  stuff(filter: "name.regex(\".*\")) {
    id
    name
  }
}
```


lukemurray:
I think it would be better to have a way to register methods you want exposed. Automatically find all available could allow people to write expressions that execute code you don't want them to etc.

PR #412, opened 2024-09-12
Add option to ignore specific data annotations

overjt:
This PR adds the ability to ignore specific data annotations when building the schema. 

Key changes:
- Added a new `IgnoreDataAnnotations` property to `SchemaBuilderOptions`
- Updated `SchemaBuilder.ShouldIncludeMember` to check for ignored annotations
- Added a unit test to verify the new functionality

This feature allows users to exclude fields with certain data annotations from the GraphQL schema, providing more flexibility in schema generation.

### DIFF
--- a/src/EntityGraphQL/Schema/SchemaBuilder.cs
+++ b/src/EntityGraphQL/Schema/SchemaBuilder.cs
@@ -319,6 +319,12 @@ namespace EntityGraphQL.Schema
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


lukemurray:
Happy to add this, just thinking about naming. Should it be `IgnoreAttributes` as they are `Attribute` classes in dotnet? Or maybe `IgnoreFieldsWithAttributes`?

Issue #441, opened 2025-01-27
Question: Mutate Arguments missing (as undefined) vs null


MarkRuseEL:
I'm working with input classes where all properties are nullable. In my mutations, which each take a single argument (args of the relevant input class), I'm finding it challenging to differentiate between properties that were explicitly set to null and those that were omitted entirely from the payload.

On the client side, I set some fields to undefined, so they are not included in the incoming payload. However, on the server, it's difficult to tell whether a property was intentionally set to null or was simply not supplied.

It would be really helpful if the library could provide a way to determine which fields were included in the payload. I've noticed that some libraries offer a feature like IResolverContext, which allows querying the fields explicitly provided in the request.  I have been through the documentation but am struggling to work out if this is supported or if there a alternative approach for handling this scenario?

lukemurray:
That is a good request. By nice to be able to have that information.

Currently don't have this, but something I'd happily add or welcome a PR.

lukemurray:
Hmm, after merging #455 and adapting it to support arguments type on query fields I realise it doesn't cover the use case in mutations if you do not use an arguments class. E.g.  your mutation is defined as

```cs
public bool MyMutation(MyContext db, int? someArg, string anotherArg) { ... }
```

There is no `someArg.IsSet()` etc.  So likely need to also (because this way works well for query fields) support having some injected service you can use in your mutation method that lets you query is an argument was set or not.

alonl153:
@lukemurray This issue is also relevant for InputTypes. From a quick check, it seems like calling the `MarkAsSet` after setting the objects value in `ProcessObjectValue` (in `QueryWalkerHelper.cs`) fixes the problem for InputTypes (and even nested InputTypes). This fix also works with GraphQLArguments even when removing the `MarkAsSet` call added in `BuildArgumentsObject` (in `ArgumentUtils.cs`). In any case, this solution should be further tested.

alon153:
> [@lukemurray](https://github.com/lukemurray) This issue is also relevant for InputTypes. From a quick check, it seems like calling the `MarkAsSet` after setting the objects value in `ProcessObjectValue` (in `QueryWalkerHelper.cs`) fixes the problem for InputTypes (and even nested InputTypes). This fix also works with GraphQLArguments even when removing the `MarkAsSet` call added in `BuildArgumentsObject` (in `ArgumentUtils.cs`). In any case, this solution should be further tested.

[...]

PR #455, opened 2025-05-20
PropertySetTrackingDto, Arguments as Dictinary<string,object?>

BRUS1NATOR:
Implementation for #441. Inherit mutation model from `PropertySetTrackingDto` so you can use `IsSet("PROPERTY_NAME")` method to check if property was provided in query.
Arguments type changed from `Dictionary<string, object>?` to `Dictionary<string, object?>?`.
I have changed `ProcessArguments` to return arguments with null value. Please check `BuildArgumentFromMember` function, it may not work correctly.

### DIFF
--- a/src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
+++ b/src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
@@ -132,7 +132,7 @@ internal sealed class EntityGraphQLQueryWalker : QuerySyntaxWalker<IGraphQLNode?
                 var directives = ProcessFieldDirectives(ExecutableDirectiveLocation.VARIABLE_DEFINITION, item.Directives);
                 foreach (var directive in directives)
                 {
-                    directive.VisitNode(ExecutableDirectiveLocation.VARIABLE_DEFINITION, schemaProvider, null, new Dictionary<string, object>(), null, null);
+                    directive.VisitNode(ExecutableDirectiveLocation.VARIABLE_DEFINITION, schemaProvider, null, new Dictionary<string, object?>(), null, null);
                 }
             }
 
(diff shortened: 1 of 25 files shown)


src/EntityGraphQL/Schema/MethodField.cs
```diff
@@ -91,17 +93,32 @@ ExecutionOptions executionOptions
...
                 allArgs.Add(argInstance);
+
+                if (p.ParameterType.BaseType == typeof(PropertySetTrackingDto))
+                {
+                    ((PropertySetTrackingDto)argInstance).MarkAsSet(gqlRequestArgs?.Keys ?? []);
+                }
             }
             else if (gqlRequestArgs != null && Arguments.TryGetValue(p.Name!, out var argField))
             {
+                //Dictionary<string, object> nonNullGqlRequestArgs = new();
```
lukemurray, line 112: Missing functionality or can be cleaned up?

BRUS1NATOR:
Removed these lines of code + improve mutation example

src/EntityGraphQL/Compiler/EntityGraphQLQueryWalker.cs
```diff
@@ -373,7 +373,7 @@ private Dictionary<string, object> ProcessArguments(IField field, IEnumerable<Ar
                 throw new EntityGraphQLCompilerException($"No argument '{argName}' found on field '{field.Name}'");
             }
             var r = ParseArgument(argName, field, arg);
-            if (r != null)
+            //if (r != null)
```
lukemurray, line 376: Also clean this up. Assuming we always add now as we handle `null` as a missing (not provided) argument

src/EntityGraphQL/Schema/MethodField.cs
```diff
@@ -91,17 +93,22 @@ ExecutionOptions executionOptions
...
+                    gqlRequestArgs ?? new Dictionary<string, object?>(),
                     Arguments.Values,
                     p.ParameterType,
                     variableParameter,
                     docVariables,
                     validationErrors
                 )!;
                 allArgs.Add(argInstance);
+
+                if (p.ParameterType.BaseType == typeof(PropertySetTrackingDto))
```
lukemurray, line 105: This is the magic (along with the `public class UpdateMovieArgs : PropertySetTrackingDto`) right?

If so, if you have another arg class that inherits from `UpdateMovieArgs` this will not work. Need to figure out if any of the base types are `PropertySetTrackingDto`.

Also we should be looking for `IPropertySetTrackingDto` right?

lukemurray:
Thanks for this @BRUS1NATOR - looks like a straightforward way to solve this. I have left a couple of comments.

PR #457, opened 2025-05-28
Added IPropertySetTrackingDto support for ObjectVariables

alon153:
Provide `IPropertySetTrackingDto` functionality for deserialization of `ObjectVariable` types (tested for InputTypes)

### DIFF
--- a/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
+++ b/src/EntityGraphQL/Compiler/Util/ExpressionUtil.cs
@@ -103,15 +103,20 @@ public static class ExpressionUtil
             if (jsonEle.ValueKind == JsonValueKind.Object)
             {
                 value = Activator.CreateInstance(toType);
+                var propSet = value is IPropertySetTrackingDto propertySet ? propertySet : null; 
                 foreach (var item in jsonEle.EnumerateObject())
                 {
                     var prop = toType.GetProperties().FirstOrDefault(p => p.Name.Equals(item.Name, StringComparison.OrdinalIgnoreCase));
                     if (prop != null)
+                    {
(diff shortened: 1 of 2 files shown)


alon153:
> I think the issue with doing it in this location, is if you use the Persisted Queries functionality (where you can send a query hash the second+ times instead of a huge graphql document) this will be cached as EGQL caches the parsed GraphQL doc, meaning when you send the hash with different variables it will be wrong.
> 
> Should be able to add a test to confirm this.
> 
> Which reminds me to add Persisted Queries to the docs.

Is this relevant for cases other than passing variables as dictionary? If an object is sent all the variables will be set with either user-defined or default values.
Added dictionary input support and tests for persisted queries receiving different dictionary inputs

lukemurray:
This issue is I was trying to describe... egql does this
1. parse GraphQL document into the `BaseGraphQLField` classes (`src/EntityGraphQL/Compiler/GqlNodes`) which hold info about the expressions etc. This may include inline variables
2. Compiles these classes into the final expression tree that is executed, combining any variables passed in the `variables` part of the JSON request
3. Executes

With Persisted Queries step 1 is not done as it has cached the objects already. Hence we don't want to be caching any variables in step 1.

However, what is happening in this code, is it's parsing inline values (those in the GraphQL query doc) which is fine to cache as that is part of the query. e.g.

```gql
{
    someField(arg1: "my-value") # OK to cache as the hash for Persisted Queries is on the whole query doc
}
```

vs.

```gql
query Q($var: String!) {
    someField(arg1: $var) # Do not cache $var value in step 1 as it can change on later requests if using Persisted Queries
}
```

All that to say I think this is fine, I'll have a better chance to look at it later though.

lukemurray:
Thanks, this fixed a few cases and I added support for when a query doesn't use variables.

PR #467, opened 2025-08-14
different approach to async

lukemurray:
Still some testing to do but the idea is
- let async things happen (don't force `GetAwaiter()`)
- clean things up at the end of it all

### DIFF
--- a/src/Benchmarks/Benchmarks.csproj
+++ b/src/Benchmarks/Benchmarks.csproj
@@ -4,14 +4,19 @@
     <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
     <IsPackable>false</IsPackable>
     <Nullable>enable</Nullable>
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.15.2" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="9.0.0">
+    <!-- Fixes package downgrade warnings (NU1608) -->
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.14.0" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="9.0.8">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="9.0.0" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="9.0.8" />
     <ProjectReference Include="../EntityGraphQL/EntityGraphQL.csproj" />
     <None Update="./DataLoader/moviedata.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs
```diff
@@ -303,14 +307,98 @@ ParameterExpression newContextParam
...
+        {
+            // Use the existing ExecuteWithConcurrencyLimitAsync method
+            return await ConcurrencyLimitFieldExtension.ExecuteWithConcurrencyLimitAsync(
+                    lambdaExpression,
+                    semaphoreConfigs,
+                    compileContext.ConcurrencyLimiterRegistry,
+                    args.Where(a => a != null).ToArray()!
+                ) ?? new object();
+        }
+        else
```
the-other-shane, line 357: else is unnessesary here (we return above)

src/EntityGraphQL/Schema/FieldExtensions/AsyncFields/ConcurrencyLimitFieldExtension.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        // 2. Service-specific limit (if specified)
+        if (serviceTypes != null)
+        {
+            foreach (var serviceType in serviceTypes)
+            {
+                var serviceLimit = executionOptions.ServiceConcurrencyLimits.GetValueOrDefault(serviceType);
+                if (serviceLimit > 0)
+                {
+                    configs.Add(($"service_{serviceType.Name}", serviceLimit));
```
the-other-shane, line 116: wondering if this will alwasy give you unique service names.

ie/ System.DateTime and MyProject.DateTime would both have Type.Name equal to "DateTime"

may also not work for generic classes. ie/ SomeRepo<TypeA>

probably worth a test, FullName might not cut it either

src/EntityGraphQL/Schema/FieldExtensions/AsyncFields/ConcurrencyLimitFieldExtension.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        return configs;
+    }
+
+    /// <summary>
+    /// Runtime execution method that applies hierarchical concurrency limiting to async operations
+    /// This gets called during expression execution, not compilation
+    /// </summary>
+    public static async Task<object?> ExecuteWithConcurrencyLimitAsync(
```
the-other-shane, line 134: have you considered additin suppirt for cancellation tokens? might be a good time to consider them. putting a note here as this is where i see them being used (so we can abort running and future async tasks if hte request itself is aborted), but it would need to be fed from the top

https://learn.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-7.0#cancellationtoken

src/EntityGraphQL/Schema/FieldExtensions/AsyncFields/ConcurrencyLimitFieldExtension.cs
```diff
@@ -0,0 +1,202 @@
...
+                // Get the result from Task<T>
+                var taskType = task.GetType();
+                if (taskType.IsGenericType)
+                {
+                    var resultProperty = taskType.GetProperty(nameof(Task<object>.Result));
+                    return resultProperty?.GetValue(task);
+                }
+                return null;
+            }
+            else
```
the-other-shane, line 171: can drop this else block

src/EntityGraphQL/Schema/FieldExtensions/AsyncFields/ConcurrencyLimitFieldExtension.cs
lukemurray: good point. Should at least be FullName. Will update.

src/EntityGraphQL/Schema/FieldExtensions/AsyncFields/ConcurrencyLimitFieldExtension.cs
lukemurray: Added in this PR.

PR #468, opened 2025-08-21
Add error path field

AndrewCasta:
Adding path field to error messages.

Starting with adding the field name of an aliased field to help with identifying error messages when processing multiple mutation fields in a single operation.

```
{
  "errors": [
    {
      "message": "Starship not found",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "path": [
        "secondShip"
      ]
    }
  ],
  "data": {
    "firstShip": "3001",
    "secondShip": null
  }
}
``` 

https://graphql.org/learn/response/#field-errors
https://spec.graphql.org/October2021/#sec-Errors.Error-result-format

### DIFF
--- a/src/EntityGraphQL/Compiler/GqlNodes/GraphQLMutationStatement.cs
+++ b/src/EntityGraphQL/Compiler/GqlNodes/GraphQLMutationStatement.cs
@@ -9,6 +9,7 @@ using EntityGraphQL.Compiler.Util;
 using EntityGraphQL.Directives;
 using EntityGraphQL.Extensions;
 using EntityGraphQL.Schema;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace EntityGraphQL.Compiler;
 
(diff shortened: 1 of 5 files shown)


src/EntityGraphQL/Compiler/GqlNodes/ExecutableGraphQLStatement.cs
```diff
@@ -82,6 +82,7 @@ QueryRequestContext requestContext
         // }
         // people & movies will be the 2 fields that will be 2 separate expressions
         var result = new ConcurrentDictionary<string, object?>();
+        var validator = new GraphQLValidator();
```
lukemurray, line 85: We should get it from the service provider. Allowing people to implement their own. But we only want to fetch it from the service provider once per mutation field - likely per top level field.

src/EntityGraphQL/Schema/MethodField.cs
```diff
@@ -84,6 +84,9 @@ ExecutionOptions executionOptions
 
         var graphQLArgumentsSet = new ArgumentsTracker();
 
+        IGraphQLValidator? validator = serviceProvider?.GetService<IGraphQLValidator>();
+        validator?.Errors.Clear(); // If the validator isn't registered as transient, we may have errors related to a different root field.
```
lukemurray, line 88: Don't think we should clear errors. We should update docs to describe how to use and register `IGraphQLValidator`. Not clearing it allows someone to register it as not transient and maybe they want to fail everything if 1 fails 🤷

PR #469, opened 2025-08-22
Make grammar immutable

sebastienros:
image

This is the flamegraph before the fix. I could see that all the time was spent in creating parse nodes and initializing their structures (ToDictionary, CharMap, ...). This is not normal, this should be done once. This meant that the parsers were begin constructed too often. The parser should be a singleton, unless some grammars vary in terms of syntax, but not behavior.

The solution is to incorporate the `ParseContext` which is specific to each invocation. And the parser lambdas can use it. This way the parser can be static. I used a local `Instance` field but I usually create my own static parser instance that I reuse. Totally fine how I did it here. Makes it more obvious for the future.


sebastienros:
I also changed the benchmarks slightly to move the initialization code out of the hot path so the results only focus on what matters. And returned the resulting expression such that the compiler would not optimize the method as the result is not used (it is not but it's a best practice for BDN usage).

PR #476, opened 2025-10-17
Extend EGQL type conversion and query capabilities; add binary comparisons and `isAny` support for custom types

samyonr:
Adding the following
- Introducing generic custom type converters (`from-to`, `to-only`, `from-only`) for flexible runtime value conversion.
- Extending EGQL with binary comparisons against non-primitive types via pluggable literal parsers (e.g., parse string literals into `Version`).
- Expanding `isAny` to support additional types through the method provider.
- Adding thorough test coverage validating the new converters, literal parsing, and combined query scenarios.

Note: `ExtendIsAnySupportedTypes` and `RegisterLiteralParser` are separate from `AddCustomTypeConverter` because we can add multiple `AddCustomTypeConverter` (to catch by from-to, to-only, from-only), and we don't want to copy and override parsing and isAny support on every "redundant" type converter - register `AddCustomTypeConverter` multiple times, for various scenarios, and  `RegisterLiteralParser` and `ExtendIsAnySupportedTypes` only once.

### Key Changes
- Schema and conversion
  - `SchemaProvider`/`ISchemaProvider`: support registering custom type converters and literal parsers.
  - `ExpressionUtil`: honors the new converters for variables/JSON-bound values.
- Query/method provider
  - `EqlMethodProvider`: extendable `isAny` supported types.
  - `Binary` grammar: support parsing string literals into target types using registered literal parsers.
- Tests
- Backward compatibility: No breaking changes expected; new behavior is opt-in via registration. I would remove the legacy type converter from the next major version

### DIFF
--- a/src/tests/EntityGraphQL.Tests/SerializationTests.cs
+++ b/src/tests/EntityGraphQL.Tests/SerializationTests.cs
@@ -174,7 +174,7 @@ internal static class JObjectTypeConverter
     }
 }
 
-internal class JTokenTypeConverter
+internal static class JTokenTypeConverter
 {
     public static object ChangeType(object value, Type toType, ISchemaProvider schema)
     {


lukemurray:
Let's remove the legacy type converter as the next version is a major version (6).

Can you expand on why `ExtendIsAnySupportedTypes` is required? Instead of `isAny` just using the default type converter in `ExpressionUtil` ?

samyonr:
Oh, I see what you did there.
Looks like it works (tests pass), and it's indeed clear.

I think I was trying not to disturb the current design too much and was worried about registering the same type multiple times (via `from-to` and `to-only`, combined). i.e., if you have both `schema.AddCustomTypeConverter<string, Version>` and `schema.AddCustomTypeConverter<Version>` (the `to-only` version, not the `from-only`).

But I think it's OK now: we have it tested in `Combining_Extensions_In_Single_Query_And_Execution_Path`. Also, `TryConvertCustom` enforces the order of converting attempts, so it's good. Previously, it was possible (potentially) to register multiple custom-converters of the same type, and the literal parser would've been registered multiple times as well - but the new approach is better.

Thanks!

Issue #484, opened 2025-12-04
Integral numeric values in filter strings parsed as long cause EF to generate bigint casts for int fields


valdecimpj:
# Description
Given a schema with nullable fields:
```
class Message
{
    public int? CustomerId { get; set; }
    public int? LeadId { get; set; }
}
```
When using filter strings like:
```
filter: "customerId == 4726"
```
EntityGraphQL initially parses integral numeric values as `long`, and its [type alignment logic for integral numeric values](https://github.com/EntityGraphQL/EntityGraphQL/blob/9476180debb0b9d2a520b43edbaa5d9845464e95/src/EntityGraphQL/Compiler/EntityQuery/Grammar/Binary.cs#L85) does not run when the left side of the comparison is a nullable `int` (since nullable types are skipped).
As a result, the right side remains of type `long` and the SQL ends up being generated as:
```
WHERE m.customer_id::bigint = 4726
```
This causes PostgreSQL to perform a type cast on the table column, which prevents index usage and results in extremely slow queries.
# Impact
- Query performance severely degraded (index not used).
- Full-table scans occur on large tables.
# Environment
EntityGraphQL version: 5.7.1
EF Core: 10.0.0
Database: PostgreSQL
.NET: 10.0

PR #485, opened 2025-12-04
Fix integral numeric types in filter expression always cast to long

valdecimpj:
https://github.com/EntityGraphQL/EntityGraphQL/issues/484


lukemurray:
Wondering if it would be better to figure out what type we are comparing to in the binary expression and make it match that type. e.g if `id` is an `int` doing `id == 3` would have `3` as an `int` not a `short`?

PR #486, opened 2025-12-04
Pass docvariables through to directives so they can use arguments

bzbetty:
Seeing exceptions when using 'include' directive in v6.

"Variable or value used for argument 'if' does not match argument type 'Boolean!' on field 'include'"


lukemurray:
Can we add a test that would fail without this change?

Issue #487, opened 2025-12-04
ResolveAsync args trying to be injected as a service


bzbetty:
```
schemaProvider.Query().AddField("field",
    new
    {
        resourceId = (int?)null
    }, "")
    .ResolveAsync<Service>((ctx, args, service) => service.blah);
```

raises exception: "Service <>f__AnonymousType57`1 not found in service provider"
from GraphQLHelper.InjectServices

weirdly the anonymous args type seems to be registered in the services list.

bzbetty:
current assumption is it's this

```
private FieldWithContextAndArgs<TContext, TParams> ResolveAsyncImpl(LambdaExpression fieldExpression, int? maxConcurrency)
{
    SetUpField(fieldExpression, true, false, true);
    Services = [.. fieldExpression.Parameters.Skip(1)];  // <--- HERE
    var serviceTypes = Services.Select(s => s.Type).ToArray();

    // if return type is Task check that it is Task<> otherwise throw
    if (typeof(Task).IsAssignableFrom(fieldExpression.Body.Type) && !fieldExpression.Body.Type.IsGenericType)
    {
        throw new EntityGraphQLSchemaException("Async field expression must return Task<T> not Task as the field needs a result.");
    }

    // Add concurrency limiting
    Extensions.Add(new ConcurrencyLimitFieldExtension(serviceTypes, maxConcurrency));

    return this;
}
```

PR #489, opened 2025-12-05
Reproduces issue #487

bzbetty:


lukemurray:
Thanks I have picked the test out without the extension changes.

PR #490, opened 2025-12-07
Refactor queryable type detection logic

bzbetty:
Replaced the use of `typeof(IQueryable).IsAssignableFrom` with a new extension method `IsGenericTypeQueryable()` in `ExpressionUtil` to determine if a type is queryable. This method now handles `Task<>` and other generic wrappers, checking recursively through interfaces for `IQueryable<>`. Updated `Expression.Call` logic to use `QueryableExtensions` or `EnumerableExtensions` based on the queryable status. Added `System.Threading.Tasks` using directive in `TypeExtensions.cs` to support the new logic.


bzbetty:
Assuming that other calls to typeof(IQueryable).IsAssignableFrom(baseExp.Type); should probably be updated too

